PR #169, opened 2015-04-03
User improvements

### LINKED ISSUE #160: Change case sensitivity for login/account creation [3]

### LINKED ISSUE #164: Allow login with email, and password reset using email [3]

beaugunderson:
This PR adds support for logging in with username or email address, makes those both case-insensitive, and enforces a case-insensitive unique constraint in the database (which requires us to use Postgres for local development instead of sqlite).

Fixes #160
Fixes #164


beaugunderson:
Going to merge this because it has tests and there's not really an alternate approach available.

PR #172, opened 2015-04-10
Upgrade to Django 1.8 and fix several related issues

beaugunderson:
I started down this path because `dumpdata` was broken after our switch to our own user class.

I ended up with this, which is still tangential work but necessary.


beaugunderson:
I've tested this locally and think it's ready for staging.

PR #178, opened 2015-04-28
Add "research" site and Researcher account

madprime:


open_humans/settings.py
```diff
@@ -274,6 +278,14 @@ def __unicode__(self):
     }
 elif dj_database_url.config():
     DATABASES['default'] = dj_database_url.config()
+else:
+    # Default to sqlite if DATABASE_URL not set.
```
beaugunderson: does this actually work?

I think this will error out if you run `./manage.py syncdb; ./manage.py migrate` right?

madprime:
oops that was a relic typo when I was messing around with things!

open_humans/settings.py
madprime: Oops - this is a relic typo of me messing around with stuff! Need to fix.

beaugunderson:
:+1: looks great except for the sqlite default; good to merge after that!

beaugunderson:
awesome, I'll get to work making this show up on staging and get my study branch ready for merge! :)

---

madprime:
awesome, nice to be making some concrete progress :)

PR #183, opened 2015-05-25
study connections UI

beaugunderson:


open_humans/settings.py
```diff
@@ -415,6 +405,27 @@ def __unicode__(self):
...
+    'social.pipeline.social_auth.social_user',
+
+    # Not needed unless we're auto-creating users
+    # 'social.pipeline.user.get_username',
+
+    # NOTE: this might be useful for UYG
+    # Associates the current social details with another user account with
+    # a similar email address.
+    # 'social.pipeline.social_auth.associate_by_email',
+    # Create a user account if we haven't found one yet.
```
madprime: Can we have a comment that indicates why we're disabling a step from the default social_auth pipeline? AIUI, it's so we force all associations to connect to an existing user – e.g. if a member logged out partway through, don't create a new user account for the token we've received.

beaugunderson:
Merging this so we can get it up on staging and fine-tune it.

PR #185, opened 2015-06-10
Make the study pages looke like Madeleine's mockups

beaugunderson:


beaugunderson:
Merging based on conversation in Slack (I showed @madprime a screenshot of the UI and then incorporated her feedback).

PR #229, opened 2015-10-03
🔧 Finish connection completion return page

### LINKED ISSUE #224: Return to OH by data-push source should be source specific URL
In the current wiki: https://github.com/PersonalGenomesOrg/open-humans/wiki/Data-connection-design-guidelines

The step in data-pull sources where the "user should be returned to Open Humans" is elaborated as going to `/member/me/research-data`. However, having data-pull sources redirect to this destination directly has two issues.

1. As detailed later in that same paragraph, for users that didn't start on Open Humans, we need to offer "a choice: return to the data source's website or remain on Open Humans". That can only be done if we're receiving them in a manner where we know the data-pull source.
2. For users that started on Open Humans, we should be able to control and change where they end up, at the end of this process. The "research data" page is our current choice, but we may discover a better design option in the future. This page might not exist in some later redesign. We might want to perform some source-specific follow-ups.

**Open Humans should receive the returned users at some other URL**, along with the origin parameter. The data-pull source should be described in this, either in the path or an additional parameter.

Then the Open Humans can display a source-specific "return or remain" page, or – if the origin indicates they started on Open Humans – redirect to the Member's research data page (our current choice for ending the process).

madprime:
Closes #224.

### DIFF
--- a/static/sass/_pages.scss
+++ b/static/sass/_pages.scss
@@ -137,6 +137,13 @@
   }
 }
 
+#connection-complete {
+  .completion-button {
+    margin-top: 1.5em;
+    margin-bottom: 1.5em;
+  }
+}
+
 #page-member-me-connections {
   div.connection {
     margin-bottom: 1em;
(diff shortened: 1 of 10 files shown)


beaugunderson:
@madprime looks great! :+1: for merging.

PR #260, opened 2015-10-24
Updates to task creation for American Gut

madprime:


studies/american_gut/signals.py
```diff
@@ -1,21 +1,27 @@
...
-def post_save_cb(sender, instance, created, raw, update_fields, **kwargs):
+@receiver(pre_save, sender=UserData)
+def pre_save_cb(**kwargs):
     """
-    Initiate retrieval of the data corresponding to an American Gut survey ID.
+    Create data retrieval task when American Gut UserData's data is updated.
     """
-    if not instance.survey_ids:
+    # Skip this signal during tests.
+    if 'test' in sys.argv:
```
beaugunderson: I think we'll want to skip this test both for fixtures and on a per-test basis but not across the board so that we can actually test the signal itself. It's possible to silence signals with decorators, and the `if raw` test from the other signals should handle the case for fixtures.

data_import/signal_helpers.py
```diff
@@ -5,6 +5,45 @@
...
+    instance = kwargs['instance']
+
+    # Require the object looks like a UserData-derived object
+    if not hasattr(instance, 'data') and hasattr(instance, 'user'):
+        return
+
+    # Only create a task if the new 'data' field is not empty and has changed.
+    if not instance.data:
+        return
+    curr_version = kwargs['sender'].objects.get(pk=instance.pk)
```
beaugunderson: Need a `try`/`except` here to handle the `DoesNotExist` case (there could be a race condition where an object is saved and then deleted in which case this signal will throw an exception)

beaugunderson:
@madprime two comments inline :)

data_import/signal_helpers.py
madprime: @beaugunderson This seems like an unlikely scenario... but another thought also occurred to me – what if UserData is newly created and unsaved (but will be) _and_ the 'data' field is not empty? In which case the correct behavior would be to go ahead (trigger a task). I've added a check on the pre_save instance's pk for this scenario.

madprime:
@beaugunderson: I guess it looks like this is triggering the task creation signal:
https://github.com/PersonalGenomesOrg/open-humans/blob/master/studies/american_gut/tests.py#L33
... and this causes an error during tests because user.member.primary_email.verified fails, as there is no associated Member in the fixture data?

data_import/signal_helpers.py
```diff
@@ -5,6 +5,55 @@
...
+    UserData object (derived from studies.models.BaseStudyUserData) and only
+    occurs if the new instance's data field is non-empty and changed.
+
+    If the user's email address is verified, the task is started. Otherwise it
+    is postponed.
+    """
+    instance = kwargs['instance']
+
+    # Skip is this is fixture data.
+    if kwargs['raw']:
```
beaugunderson: any kwarg that we always use could be added to the method parameters directly (so this one and `sender`)

beaugunderson:
I added the missing fixture to make the tests pass. :)

:+1: to merge; you can change the kwargs to method parameters if you want but I'm ok to merge without

PR #279, opened 2015-11-02
Updates to welcome

madprime:


open_humans/templates/member/welcome-data-import.html
```diff
@@ -69,16 +80,43 @@
       Import your data from other data sources &ndash; to share with other
       members, studies, or even publicly!
     </p>
-
     <p>
-      This is a feature we're working to add, and it will evolve and grow. What
-      data sources would you like us to support? What we build is largely
-      guided by what our Members want to contribute!
+      You'll be able to share this data sets directly with individual studies.
```
beaugunderson: typo: "share this data sets"

open_humans/templates/partials/import-button.html
```diff
@@ -2,7 +2,7 @@
   {% if user.member.primary_email.verified %}
     {% if source %}
     <a class="btn btn-primary {{ class }}"
-      href="{{ source.href_connect }}?origin=open-humans&next={{ source.href_next }}">
+      href="{{ source.href_connect }}?origin=open-humans&amp;next={{ source.href_next }}">
```
beaugunderson: @madprime this seems incorrect (or if this change is required it seems like something else is broken)

open_humans/tests.py
```diff
@@ -10,13 +10,13 @@
...
 
 ANONYMOUS_URLS = [
+    '/',
     '/account/login/',
     '/account/password/reset/',
     '/account/signup/',
 ]
 
 AUTHENTICATED_OR_ANONYMOUS_URLS = [
-    '/',
```
beaugunderson: @madprime `/welcome/profile/` should be added to `AUTHENTICATED_URLS` so it can be tested :)

beaugunderson:
@madprime some comments inline but otherwise :+1: :)

open_humans/templates/partials/import-button.html
madprime: I changed it because the HTML linting my Atom was using flagged it. This seems to indicate it's correct? http://stackoverflow.com/questions/3705591/do-i-encode-ampersands-in-a-href

PR #280, opened 2015-11-05
Add welcome email

beaugunderson:
Take a quick look @madprime?

### DIFF
--- a/open_humans/templates/email/welcome.html
+++ b/open_humans/templates/email/welcome.html
@@ -18,7 +18,8 @@
   Sharing Data Study turns on the “public sharing” feature. You’ll be able to
   turn “public” status on or off for each of your data sets.</p>
 
-<p>Get started by visiting our Welcome Page.</p>
+<p>Get started by visiting our <a href="{{ welcome_page_url }}">Welcome
+  page</a>.</p>
 
 <p>With your help, we’ll start breaking down data silos in human health and
   research.</p>
--- a/open_humans/templates/email/welcome.html
+++ b/open_humans/templates/email/welcome.html
@@ -14,9 +14,10 @@
   with, and studies that will share data with you. Build a profile that
   connects data across studies!</p>
 
-<p>You can even choose to make some data sets public. Joining our Public
-  Sharing Data Study turns on the “public sharing” feature. You’ll be able to
-  turn “public” status on or off for each of your data sets.</p>
+<p>You can even choose to make some data sets public. Joining our <a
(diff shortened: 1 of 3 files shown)


open_humans/templates/email/welcome.html
```diff
@@ -0,0 +1,33 @@
...
+  need to sign up for “news and updates” on your account settings. We really
+  don’t email much! We email to tell Members about new studies and activities,
+  and give updates about progress in current studies.{% endif %}</p>
+
+<p>We believe research should go both ways: find studies to join and share data
+  with, and studies that will share data with you. Build a profile that
+  connects data across studies!</p>
+
+<p>You can even choose to make some data sets public. Joining our Public
+  Sharing Data Study turns on the “public sharing” feature. You’ll be able to
```
madprime: We should link Public Data Sharing

open_humans/templates/email/welcome.txt
```diff
@@ -0,0 +1,34 @@
...
+
+We believe research should go both ways: find studies to join and share data
+with, and studies that will share data with you. Build a profile that connects
+data across studies!
+
+You can even choose to make some data sets public. Joining our Public Sharing
+Data Study turns on the “public sharing” feature. You’ll be able to turn
+“public” status on or off for each of your data sets.
+
+Get started by visiting our Welcome page[1].
```
madprime: Hm... limitations of text. I think it's better to do: `Get started by visiting our Welcome page: {{ welcome_page_url }}`

Also, above:
`Joining our Public Data Sharing study turns on the "public sharing" feature: {{ public_data_url }} You'll be able to turn "public" status on or off for each of your data sets.`

open_humans/templates/email/welcome.txt
```diff
@@ -0,0 +1,34 @@
+Thank you for joining the Open Humans community!
+
+To make sure you get future emails from us, add our email to your “contacts”.
+
+{% if newsletter %}You’ve signed up for news and updates, but without this
+step our emails might be “filtered” by your email client.{% else %}
```
madprime: I think I'm seeing non-ASCII characters (left and right quotation marks). Would this be an issue, or is non-ASCII fine? (Note, non-ASCII quote marks are also in the HTML email.)

beaugunderson: I considered replacing them with the standard equivalents but they did come through in the email (and look nicer). We can deploy to staging and you can sign up for a new account to get the welcome email and verify that they work for you too?

PR #282, opened 2015-11-09
Wildlife of Our Homes study

beaugunderson:


madprime:
are the browserstack binaries meant to be in our repository?

beaugunderson:
@madprime yes, it's the easiest way to distribute them since they're used both for local development and for CI; ideally they'd be in `pip` or `npm` or some other package manager but because they're not I included them. :)

studies/wildlife/templates/scopes/wildlife.html
```diff
@@ -0,0 +1,19 @@
+{% extends 'scopes/scope-container.html' %}
+
+{% block icon %}
+<span class="glyphicon glyphicon glyphicon-folder-open"
+  aria-hidden="true"></span>
+{% endblock %}
+
+{% block description %}
+<p>This permission grants access to your <strong>Wildlife of Our Homes</strong>
```
madprime: Are we still using this material? I thought our rewrite of the interactions removed this stuff.

PR #283, opened 2015-11-09
PGP interstitial

beaugunderson:


open_humans/urls.py
```diff
@@ -86,6 +86,8 @@
         name='terms-of-use'),
     url(r'^activities/$', ActivitiesView.as_view(), name='activities'),
     url(r'^statistics/$', StatisticsView.as_view(), name='statistics'),
+    url(r'^pgp-interstitial/$', PGPInterstitialView.as_view(),
```
madprime: Let's call it 'pgp-quick-note'

open_humans/middleware.py
```diff
@@ -78,3 +81,28 @@ def process_request(request):
...
+    def process_request(request):
+        if request.user.is_anonymous():
+            return
+
+        if request.user.member.seen_pgp_interstitial:
+            return
+
+        public = (request.user.pgp.datafile_set
+                  .filter(_public_data_access__is_public=True))
+        private = (request.user.pgp.datafile_set
```
beaugunderson: use `is_enrolled` here instead

open_humans/middleware.py
```diff
@@ -78,3 +81,28 @@ def process_request(request):
...
         return get_production_redirect(request)
+
+
+class PGPInterstitialRedirectMiddleware:
+    """
+    Redirect users with more than 1 private PGP datasets and zero public
+    datasets to an interstitial page exactly one time.
+    """
+    @staticmethod
+    def process_request(request):
```
madprime: add return if not a PGP user

open_humans/middleware.py
```diff
@@ -78,3 +81,28 @@ def process_request(request):
...
+
+        if request.user.member.seen_pgp_interstitial:
+            return
+
+        public = (request.user.pgp.datafile_set
+                  .filter(_public_data_access__is_public=True))
+        private = (request.user.pgp.datafile_set
+                   .filter(_public_data_access__is_public=False))
+
+        if private.count() > 0 and public.count() < 1:
```
madprime: if public.count > 0 ... skip interstitial and change seen_pgp_interstitial to True

PR #284, opened 2015-11-09
23andMe file upload

beaugunderson:


activities/twenty_three_and_me/models.py
```diff
@@ -8,6 +6,14 @@
 from data_import.models import BaseDataFile, DataRetrievalTask
 
 
+def get_upload_path(instance, filename=''):
+    """
+    Construct the upload path for a 23andMe upload.
+    """
+    return 'member/{}/uploaded-data/23andme/{}'.format(instance.user.id,
```
madprime: Idea - maybe timestamp so we don't overwrite files with the same name, from the same user? thoughts?

activities/twenty_three_and_me/signals.py
```diff
@@ -0,0 +1,21 @@
...
+
+from .models import UserData, DataFile
+
+
+@receiver(post_save, sender=UserData)
+def post_save_cb(instance, created, raw, **kwargs):
+    """
+    Create data retrieval task when 23andMe UserData's data is updated.
+    """
+    if not instance.file_url:
```
madprime: Maybe do this pre_save and check if it's changed, as we've done for the new American Gut UserData?

PR #294, opened 2015-11-20
move subtype to base model

beaugunderson:
figured out a way to do this by keeping the defaults in the models as a field and overriding `save` in the base, but only after futzing around with using a callable that was part of the class (which doesn't work).


madprime:
Looks good to me

PR #296, opened 2015-11-20
centralize subtypes and their display

beaugunderson:


madprime:
Looks great to me :+1:

PR #300, opened 2015-11-24
Fix broken task trigger, do this in view instead of signal

madprime:


open_humans/storage.py
```diff
@@ -4,7 +4,6 @@
 
 TEN_MINUTES = 60 * 10
 
-
```
beaugunderson: from PEP8: "Surround top-level function and class definitions with two blank lines."

beaugunderson:
@madprime one comment, otherwise :+1:!

PR #308, opened 2015-12-07
Changes to research data page (stage 1)

beaugunderson:


data_import/models.py
```diff
@@ -44,7 +46,29 @@ class DataRetrievalTaskQuerySet(models.QuerySet):
...
+        most recent dataset if not.
+        """
+        with_files = [d for d in datasets if d.data_files]
+
+        if with_files:
+            return with_files[0]
+
+        return datasets[0]
+
+    def grouped_recent(self):
```
madprime: Let's add a docstring that explains what this returns. I think it's a dict where the key is the 'source' field and the value is the DataRetrievalTask returned by the "most_recent" method for all DataRetrievalTasks for this source.

PR #316, opened 2015-12-07
Public sharing by source

madprime:


beaugunderson:
closing in favor of #317--I know I said it didn't matter but I found it too hard to review the diff with both sets of changes in it so I created a new PR against bg-research-data so I could see just the diff of your changes. :)

data_import/models.py
```diff
@@ -277,16 +297,37 @@ def __unicode__(self):
         return '%s:%s:%s' % (self.user_data.user, self.source, self.file)
 
     # This is the inverse relation of the GenericForeignKey defined in the
-    # PublicDataAccess model.
-    _public_data_access = GenericRelation(PublicDataAccess,
-                                          content_type_field='data_file_model',
-                                          object_id_field='data_file_id')
+    # DataFileAccessLog model.
+    datafileaccesslog_set = GenericRelation(
```
beaugunderson: I think this should be renamed to `logs` or `access_logs`

open_humans/views.py
```diff
@@ -695,53 +701,54 @@ def get_inbound_connections():
...
-            app_label = f.pop('data_file_model__app_label')
-            app = app_from_label(app_label)
-
-            f['app'] = app.verbose_name
-
-        # sort here instead of in the database since we need to look up the
-        # Django app name
-        return sorted(files, key=itemgetter('app'))
+    def get_source_connections(self):
+        source_connections = dict()
```
beaugunderson: I think the idiom is to prefer `{}` vs. `dict()` for empty dicts (not that it matters much!)

beaugunderson:
looks good so far; going to try running it locally now

beaugunderson:
works locally as well--going to add some status updates to the migration but we could probably test on staging :)

open_humans/views.py
madprime: Hadn't realized there was a speed difference! Microscopic, of course. Sadly not what is causing this code to be slow as hell. I'm sure that "iterating over each UserData, each time performing a filter on PublicDataAccess objects" was not a brilliant approach. Surely a better solution exists, but at this point I just wanted to fix the page - albeit incompletely - so I could move on.

madprime:
I added some style stuff to the research data page. I wanted more visual clarity on what source each public data sharing button is associated with, an unrelated task but I'm much happier with how it looks now...

PR #317, opened 2015-12-08
WIP: Public sharing by source

beaugunderson:


beaugunderson:
@madprime looks good so far; one suggestion is to rename `public_data_access` to something like `is_shared` or `is_public` since that's a more fitting name for a boolean property, I think...

(and it will also flush out any consumers of `public_data_access` that might break if a boolean is now returned)

PR #326, opened 2015-12-15
Django 1.9 support

beaugunderson:
Tested locally with production data.

@madprime can you also test when you have time? not high priority, but finished this PR up because django-hosts released a new version (which was the only thing holding up 1.9 support)


madprime:
Looks good to me! :+1:

PR #342, opened 2016-01-04
Improve source and files display

madprime:


open_humans/views.py
```diff
@@ -69,7 +69,7 @@ def get_context_data(self, **kwargs):
         context.update({
             'next': reverse_lazy('member-detail',
                                  kwargs={'slug': self.object.user.username}),
-            'public_data': self.object.public_data_participant.public_data,
+            'public_data_tasks': self.object.public_data_participant.public_data_tasks,
```
beaugunderson: > 80 characters

beaugunderson:
@madprime 1 nit, otherwise looks good :+1:

PR #344, opened 2016-01-07
Add source data file deletion

beaugunderson:


data_import/templatetags/data_import.py
```diff
@@ -0,0 +1,20 @@
...
+
+from ..utils import app_name_to_user_data_model
+
+register = template.Library()
+
+
+@register.simple_tag
+def source_is_connected(source, user):
+    """
+    Given 'go_viral', return True or False.
```
madprime: This description seems a bit rough :)
(although it does seem clear from the name of the tag)

open_humans/templates/member/my-member-research-data.html
```diff
@@ -1,89 +1,103 @@
...
-        <input type="hidden" name="source" value="{{ source }}">
+    <div class="pull-right">
+      {% if task.data_files %}
+        {% source_is_connected source user as is_connected %}
 
-        {% if task.is_public %}
-          <input type="hidden" name="public" value="False">
-        {% else %}
-          <input type="hidden" name="public" value="True">
+        {% if is_connected %}
```
madprime: I thought we only offer delete if it's not connected?

open_humans/tests.py
```diff
@@ -169,39 +173,39 @@ def test_lowercase_unique(self):
...
-#            'testing123')
-#
-#        driver.find_element_by_name('terms').click()
-#
-#        driver.find_element_by_id('create-account').click()
-#
-#        self.assertEqual(
-#            'Please verify your email address',
-#            driver.find_element_by_css_selector('h3.panel-title').text)
+# class OpenHumansBrowserTests(BrowserTestCase):
```
madprime: Are we holding on to this because it might be usable again at some point? Maybe add a comment on why the commented code is retained but commented out?

open_humans/views.py
```diff
@@ -283,15 +280,36 @@ def get_context_data(self, **kwargs):
...
     def get_queryset(self):
-        return DataRetrievalTask.objects.filter(user=self.request.user)
+        source = self.kwargs['source']
+
+        data_file_model = app_name_to_data_file_model(source)
+
+        return data_file_model.objects.filter(
+            user_data__user=self.request.user)
+
+    def get_object(self, queryset=None):
```
madprime: Hah, this looks like a hack... but because queryset also has a delete() method I guess it works?

To be safe, I wonder if it's better to do it explicitly rather than trusting DeleteView to always have the right behavior. Not a big deal, maybe just make a comment in the code noting it's a hack (unless I'm wrong about that)

beaugunderson: not really a hack, get_queryset is designed to be overridden to get the queryset for deletion; the only quirk is that DeleteView expects to then filter the queryset by the pk, which is done in get_object (it expects one object). since we pre-filter in get_queryset we just pass that through. it is probably more readable to just get the queryset in get_object though! that way we only override one method.

open_humans/templates/member/my-member-research-data.html
beaugunderson: oops yes, this went through a couple of iterations and the `not` got lost, good catch :)

data_import/templatetags/data_import.py
beaugunderson: true, I just updated the description from the tag it was based on--I'll add a better description. :)

open_humans/views.py
madprime: But as a result, self.object is not actually an object, it's a queryset. It works because calling delete() on a queryset also works? But it seems like the language of DeleteView anticipates an object: https://github.com/django/django/blob/master/django/views/generic/edit.py#L257

madprime:
Looks good to me! :+1:

PR #349, opened 2016-01-11
Allow removal of datafiles during disconnection

beaugunderson:


madprime:
Looks good to me! :+1:

PR #351, opened 2016-01-11
Add timestamp to task S3 key "directory" and fix naive datetime

madprime:


beaugunderson:
@madprime looks good to me; I wish Django had named `timezone.now` `utc.now` to reflect what it actually does though :)

PR #363, opened 2016-01-13
Data selfie

beaugunderson:


open_humans/templates/member/my-member-data-selfie.html
```diff
@@ -0,0 +1,157 @@
...
+</div>
+{% endblock %}
+
+{% block dashboard_main %}
+<div class="panel panel-default pad-all-sides">
+  {% with source='data_selfie' %}
+    {% source_is_public source as is_public %}
+    {% source_is_connected source user as is_connected %}
+
+    {% include 'partials/share-and-delete-buttons.html' %}
```
madprime: The public sharing button, as included on this page, directs back to research data when someone clicks "share publicly". I think the user should "stay" (or actually stay, with AJAX?) on this page when they update public status on this button.

madprime:
Just had one bump in user experience I commented on, otherwise :+1: I look forward to seeing this is staging!

open_humans/templates/member/my-member-data-selfie.html
beaugunderson: ah yes, the redirect there is unintentional, it should be using AJAX—I’ll investigate

beaugunderson: oh, needs the JavaScript included on that page, simple fix :)

PR #364, opened 2016-01-14
Move from sass to PostCSS

beaugunderson:
My OS X environment broke today, leaving me unable to compile libsass.

There's a lot of support for PostCSS nowadays, and it supports sass nesting and variables.

It also supports the actual future CSS syntax that browsers will support, which is nice. Most importantly, it doesn't require a compiled libsass, so builds and deploys get faster (and break less frequently; we've had a couple of issues with both codeship and heroku and libsass).

So with that in mind I switched us over to postcss; the changes from sass are:

- `/* */` comments are the only type supported
- use the `color()` function instead of `darken()` and `lighten()` (this is part of the CSS spec)
- use the `calc()` function for arithmetic (also part of the CSS spec)
- `@import` files by full name

I also added CSS minimization via cssnano and fixed up some older issues in our gulpfile, as well as migrated us from minifyify to uglify2 for JS minimization (faster in my testing, and seems to be the standard of the gulp community).

I removed the `bundle-` prefix from the generated files as well so the filenames are 1:1 now.


madprime:
looks good to me! :+1:

PR #366, opened 2016-01-15
move from gunicorn to uWSGI

beaugunderson:


open_humans/uwsgi.ini
```diff
@@ -0,0 +1,17 @@
...
+http-socket = :$(PORT)
+lazy = true
+master = true
+max-requests = 5000
+memory-report = false
+module = open_humans.wsgi:application
+processes = 4
+
+; If VIRTAL_ENV is set then use its value to specify the virtualenv directory
+if-env = VIRTUAL_ENV
```
madprime: Add to env.example?

beaugunderson: this is managed automatically by `virtualenv`; whenver you have a virtualenv activated (as you will after `workon <virtualenv-name`) it will be set :)

madprime:
looks good to me! :+1:

PR #367, opened 2016-01-15
Add webhook alert for new member signups

beaugunderson:
these would appear in the `#oh-analytics` Slack channel.


madprime:
This looks good to me! :+1:

The addition of `account.user.member = member` looks like a wise approach to be making, given how we've hit bugs like that in the past! :)

PR #368, opened 2016-01-16
Add hosted Piwik tracking

beaugunderson:
This pull request would add the Piwik tracking code to Open Humans.

Piwik is a [privacy-conscious](http://piwik.org/privacy/) analytics company; their main offering is open source and can be self-hosted but I think it's worth paying for them to host it ($29.95/month).

I think we're at a size where having insight into who's coming to the site and where we're getting traffic from will be useful.

Thoughts @madprime? You can login via Meldium to see the interface, we have a 30 day trial.

### DIFF
--- a/open_humans/templates/base.html
+++ b/open_humans/templates/base.html
@@ -137,6 +137,7 @@
 
     {% block extra_modals %}{% endblock %}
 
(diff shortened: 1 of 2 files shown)


madprime:
let's go ahead and try it out! :+1:

PR #369, opened 2016-01-20
Add <title> tags to every page

beaugunderson:
This PR adds a `head_title_suffix` that defaults to " - Open Humans" and changes all uses of `panel_title` to `head_title`. It also specifies titles where no title was previously specified.

For titles that contain "Open Humans" like "Log in to Open Humans" I set `head_title_suffix` to an empty string.

(I noticed that the `<title>` of every page was set to "Open Humans")

### DIFF
--- a/open_humans/templates/account/delete.html
+++ b/open_humans/templates/account/delete.html
@@ -2,7 +2,7 @@
 
 {% load utilities %}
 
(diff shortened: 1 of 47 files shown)


madprime:
:+1: Looks good to me!

PR #395, opened 2016-02-11
Make data selfie show up on public profile

madprime:
@beaugunderson I ended up making a fix that I strongly suspect will be obviated by your refactoring work, but we want it fixed now, so I went ahead. I've opened a PR so you can see what's happening here. It's EOD for me, please go ahead and merge & deploy to production if you think it's OK.


beaugunderson:
looks good to me (and yes, the refactor will also simplify this) :)

PR #407, opened 2016-02-16
Upload improvements

madprime:
Remove hard-coded upload paths and improve wording on the 23andMe upload form.


activities/data_selfie/views.py
```diff
@@ -18,12 +19,11 @@ class UploadView(PrivateMixin, DropzoneS3UploadFormView):
...
     def get_upload_to(self):
-        return ('member/{}/uploaded-data/data-selfie/{}/'
-                .format(self.request.user.id, int(time())))
+        return get_upload_dir(DataSelfieDataFile, self.request.user)
 
     def get_upload_to_validator(self):
-        return (r'^member/{}/uploaded-data/data-selfie/\d+/'
-                .format(self.request.user.id))
+        return (r'^member/{}/uploaded-data/{}/\d+/'.format(
+            self.request.user.id, DataSelfieDataFile._meta.app_label))
```
beaugunderson: let's make this into a `get_upload_dir_validator` with the same arguments as `get_upload_dir`?

activities/twenty_three_and_me/models.py
```diff
@@ -28,6 +18,9 @@ class Meta:
     user = fields.AutoOneToOneField(settings.AUTH_USER_MODEL,
                                     related_name='twenty_three_and_me')
 
+    # As of Feb 2016, we believe the the upload_to defined below is always
+    # overridden by the DropzoneS3UploadFormView upload_to definition.
+    # Maintaining here in case we use the version defined below in the future.
     genome_file = models.FileField(upload_to=get_upload_path, max_length=1024,
```
beaugunderson: anything we're only keeping to use in the future I would just remove because it will always be in git history :)

activities/twenty_three_and_me/views.py
```diff
@@ -22,12 +21,11 @@ class UploadView(PrivateMixin, DropzoneS3UploadFormView, DataRetrievalView):
...
     def get_upload_to(self):
-        return ('member/{}/uploaded-data/data-selfie/{}/'
-                .format(self.request.user.id, int(time())))
+        return get_upload_dir(self.model, self.request.user)
 
     def get_upload_to_validator(self):
-        return (r'^member/{}/uploaded-data/data-selfie/\d+/'
-                .format(self.request.user.id))
+        return (r'^member/{}/uploaded-data/{}/\d+/'
+                .format(self.request.user.id, self.model._meta.app_label))
```
beaugunderson: same as above

beaugunderson:
some small comments, good to merge after they're addressed :+1:

activities/twenty_three_and_me/models.py
madprime: I'm worried about a lack of omniscience here. If this is merely left to history and no comment is here, then if Dropzone is removed in the future someone may not realizing the setting needs to be added back in. The way it is now, if someone removes Dropzone, it's robust - it continues to behave correctly.

madprime:
@beaugunderson I didn't really address one... I'm reluctant to remove the robustness of retaining upload_to as consistently defined in case we somehow receive non-Dropzone file uploads in the future. (And I don't expect someone to realize "it's in git history".) I trimmed the wordiness of it, though. Do you care strongly?

beaugunderson:
sorry, my comment re: git history was based on a misunderstanding; I thought we were going to keep the `get_upload_path` definition in that file (which wasn't case, I just missed its removal). that's what I meant when I said it would be in git history if we needed it again, totally OK with the comment & :+1: to merge :)

PR #421, opened 2016-02-29
Refactoring for activities

beaugunderson:
lots of refactoring specifically to make adding new UserSocialAuth-based activities very easy; plus some template cleanup and a bug fix or two


activities/app_configs.py
```diff
@@ -0,0 +1,42 @@
...
+class UserSocialAuthAppConfig(BaseConnectionAppConfig):
+    """
+    An AppConfig used for activities that connect via UserSocialAuth.
+    """
+
+    def get_user_data(self):
+        outer_self = self
+
+        @property
+        def get_user_data_inner(self):
```
madprime: Trying to wrap my head around this wrapping... Why was it wanted/needed/useful? Also, does `user=self` mean a UserSocialAuthUserData object is created with this UserSocialAuthAppConfig object as the `user` parameter?

activities/user_data.py
```diff
@@ -0,0 +1,58 @@
...
+        self.user.social_auth.filter(provider=self.provider).delete()
+
+    def get_retrieval_params(self):
+        return {
+            'access_token': self.get_access_token(),
+        }
+
+    def get_access_token(self):
+        """
+        Get the access token from the most recent RunKeeeper association.
```
madprime: Copy pasta? :)

activities/app_configs.py
beaugunderson: @madprime Sorry, this should have been documented better :)

The `UserSocialAuthAppConfig`'s `ready` method adds an attribute to the User model with the name set to the app's `label`.

So in the case of RunKeeper, User gets an attribute called `runkeeper`. The wrapping is to provide that attribute the correct `User` when it's instantiated.

This lets us maintain the old pattern of accessing this data (and provides parity with the way study data is accessed from the `User` object) but makes it so we don't actually have to create a model for activities that use `UserSocialAuth` for their connection (because the model would not contain any additional fields, so it would just be database clutter).

activities/user_data.py
beaugunderson: oops yes!

madprime:
Fixed a bug in deauthorization for RunKeeper, and added a test for what broke. Nicer to have a method shared by SourceDataFilesDeleteView (in open_humans/models.py), but we shouldn't let perfect be the enemy of good enough! (I think this is good enough.) :+1:

PR #427, opened 2016-03-04
move data-selfie templates/views

beaugunderson:


madprime:
Looks like some sensible refactoring, I don't see any issues looking it over visually (but didn't actually run it to test locally)! :+1:

PR #439, opened 2016-03-10
fix ubiome file model

madprime:
The DataFile model is only for files intended for return/display.
The only exception for exposing raw user uploads is Data Selfie.
All other DataFile cases are the result of our data process, not
original uploads.


beaugunderson:
:+1:

PR #448, opened 2016-03-23
Add support for OAuth2 projects

beaugunderson:


beaugunderson:
@madprime: for your perusal :)

PR #492, opened 2016-05-23
Work for improving auth templates

beaugunderson:


beaugunderson:
@madprime ready for feedback I think :)

madprime:
:+1:  I think this looks good to go!

PR #494, opened 2016-05-31
Improve project authorization panel

madprime:


beaugunderson:
@madprime looks good to me! 👍

PR #518, opened 2016-06-14
Check Illumina UYG file name and size upon upload

madprime:


beaugunderson:
@madprime: looks good, 👍!

PR #519, opened 2016-06-14
uBiome form update

madprime:


beaugunderson:
@madprime: looks good, 👍!

PR #539, opened 2016-07-05
Activities → home

madprime:
Also removing the "welcome" pages.


open_humans/templates/pages/home.html
```diff
@@ -1,87 +1,21 @@
...
-            <div class="col-sm-4">
-              <p>
-                <a class="btn btn-primary btn-sm"
-                    href="{% url 'my-member-settings' %}">
-                  Account settings
-                </a>
-              </p>
+{% block extra_js %}
+<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery.imagesloaded/4.1.0/imagesloaded.pkgd.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery.isotope/2.2.2/isotope.pkgd.min.js"></script>
```
beaugunderson: I think we'll want to add these to package.json and include them locally so we can see when they become outdated (but this can happen later)

beaugunderson:
@madprime looks good! 👍

PR #548, opened 2016-07-16
Fix direct sharing issues

madprime:


madprime:
( @beaugunderson said fix looked good in slack )

Issue #566, opened 2016-08-08
Add Twitter and Facebook links to footer


madprime:

jasonbobe:
thanks @madprime

PR #569, opened 2016-08-08
Edit base.html to fix Header and Footer: Facebook and Twitter add PNGs

BenjaminHCCarr:
Added two PNGs in static: Facebook and Twitter from media packs
Added Activities to Header
Added Twitter to Facebook to Footer

Addresses #566


open_humans/templates/base.html
```diff
@@ -127,6 +131,8 @@
           <a href="{% url 'ed-recruitment' %}">ED&nbsp;position</a> |
           <a href="http://blog.openhumans.org/">Blog</a> |
           <a href="http://forums.openhumans.org">Forums</a> |
+          <a href="https://twitter.com/openhumans"<img src="{% static images/twitter-xs-logo.png' %}" alt="Twitter">@OpenHumans</a> |
```
beaugunderson: missing a closing &gt; here

open_humans/templates/base.html
```diff
@@ -127,6 +131,8 @@
...
+          <a href="https://www.facebook.com/openhumansorg/"<img src="{% static 'images/FB-f-Logo__blue_29.png' %}" alt="Facebook">@openhumansorg</a> |
```
beaugunderson: also missing the closing &gt; here

BenjaminHCCarr:
These are fixed.
This was actually the need for the edit in #7953999 where I had the double ">>" on those same lines.
I had not split them properly when I added the images to the links.

open_humans/templates/base.html
```diff
@@ -76,7 +76,11 @@
             <li class="{% active 'research' %}">
               <a href="/research/">Research</a>
             </li>
-
+            
+            <li class="{% active 'activities-list' %}">
+              <a href="{% url '#activities-list' %">Activites</a>
```
madprime: I removed 'Activities' from the navbar when the list moved on to the 'home' page...

Is it a UX thing, users can't find their way back to the link on the home page after exploring other pages? (This makes sense to me. It hadn't occurred to me.)

Also, I haven't checked, but the code presumably barfs on line -- it's trying to reference a [named URL pattern](https://docs.djangoproject.com/en/1.10/topics/http/urls/#naming-url-patterns) that doesn't exist (i.e. '#activities-list' that isn't named in /open_humans/urls.py)

I think the correct link for this should be: `href="{% url 'home' %}#activities-list"`

BenjaminHCCarr:
hmmm. I assume '--' is line '81', based on your link, their example is similar, I also based my use of the  '{% url' syntax on the other URLs in the page, possibly it behaves differently because of the leading hash? I don't have a django setup to test on, and would need to clone everything and am remote today anyway.

```
The example they use is
According to this design, the URL for the archive corresponding to year nnnn is /articles/nnnn/.
<a href="{% url 'news-year-archive' 2012 %}">2012 Archive</a>
```

I was confused. I now see in  [urls.py you have defined 'home'](https://github.com/PersonalGenomesOrg/open-humans/blob/master/open_humans/urls.py#L78) which I didn't know, as I didn't see it referenced in the rest of base.html:
`url(r'^$', views.HomeView.as_view(), name='home')`

so you are right the correct edit would likely be:
`<a href="{% url 'home' %}#activities-list">`

Which is actually how you committed it in [about-main-content-2.html#L21](https://github.com/PersonalGenomesOrg/open-humans/blob/51569169a8587514e6ddec70c831d0d26f203f63/open_humans/templates/pages/about-main-content-2.html#L21) (still trying to understand django)

I understand wanting a clean UX, but, I wasn't sure about how to find the activities page, it never occurred to me to go "home" to look for things. I kept clicking on "Research" and wasn't sure why that was featured in the sidebar for a logged in user, when 99.y% of users will never need to decide between Oauth2 or not for their project, or even launch a project.

I love the [https://www.openhumans.org/#filter=](https://www.openhumans.org/#filter=) stuff, but I wouldn't think to go their from my "Me"/"My Account" Pages, and was actually looking in the header and sidebar for a link, hence my edits.

madprime:
I'm closing this in favor of #572, which accepts all your changes and adds some fixes. Thank you!

To follow on this thread: yeah, I was using the Github interface to comment on line 81.

If you're interested in doing more work on the website's code, we'd be happy to help you get set up for local development! I'd like to send you an invitation to one of our Slack channels (we've invited other contributors/engaged coders to the channel -- but you're the first looking at the website code proper!). DM me what email to send the Slack invite on Twitter (@madprime), if you're up for that.

Adding "Activities" link back into the navbar sounds like an improvement, let's do it -- and thank you for pointing out the usability problem here! Your points are spot on.

I do think we need the "Research" tab right now ... at least for anonymous (not logged in). We're working really hard right now to increase researcher awareness and interest.

But you're right. For a logged in member, it's extremely unlikely to be relevant. We should think about how to bury it. Some users will be researchers -- I'm not sure what's best, I want to think about that more.

For now, hopefully adding "Activities" back in to the navbar is a good enough improvement, and we'll think about how to further improve it. Thank you!

Issue #570, opened 2016-08-09
'Upload Data Selfie' Completes without Hitting Continue


BenjaminHCCarr:
When using the [](https://www.openhumans.org/activity/data-selfie/upload/) option, any file added to the box either by drag-and-drop or selection is uploaded without hitting:

> Upload a file before continuing

This was unexpected behavior as I selected the wrong file to upload with the file selection tool, and hit the browser back button, and went forward again to upload the correct file as there was no removal option I could find, but the file still ended up in my 'activities:data-selfie' pool.

madprime:
@BenjaminHCCarr do you think this is resolved by #571?

BenjaminHCCarr:
I think it makes things clearer and I understand the desire to start moving files and the UX experience, but I'm also old school, I would prefer a queue that goes on clicking continue, but that is a separate issue and more up to you. I also understand that the current method allows uploads while the user picks more objects, I was just surprised by the behavior and that #571 does tell the user what is going to happen and your merge that you can delete on next page helps as well. Again it's what the OH team wants, this is a WEB 2.something-something thing. But I would much rather see time spend on coarser management/sharing of selfies, and time spent on editing the upload feature as it works well, and it's behavior is better documented and the way to delete is clear thanks to your merge.

madprime:
Thanks! I'm going to "freezer" this as low priority without closing it.

Because the user is uploading immediately and directly to S3 using JS (and notifying Open Humans when the upload is complete), it's more complex to manage files in an "intuitive" manner. But not impossible.

PR #571, opened 2016-08-09
Update data_selfie/upload.html to make clear upload automatic

BenjaminHCCarr:
This clarifies the behavior of the data_selfie system, and file dropped or selected is uploaded automatically, even if you don't hit "continue".

ref issue #570


madprime:
:+1: LGTM!

@beaugunderson, merge at will

PR #572, opened 2016-08-09
[Updated] Edit base.html to fix Header and Footer: Facebook and Twitter add PNGs

madprime:


madprime:
I've created a new PR with some fixes added -- thanks @BenjaminHCCarr !

PR #580, opened 2016-09-01
New activity pages

beaugunderson:
All of the context is in the view and it's also displayed on the page but the layout needs some visual work and some wordsmithing.


open_humans/templates/member/activity-management.html
```diff
@@ -0,0 +1,246 @@
...
+                {{ data_file.access_logs.count }}
+              </td>
+
+              <td>
+                {{ data_file.description }}
+              </td>
+
+              {% if is_individual_deletion %}
+              <td>
+                <a href="#" class="btn btn-danger btn-xs">Delete</a>
```
madprime: Is this button doing anything?

PR #583, opened 2016-09-09
Migrate GoViral and associated files to a project

beaugunderson:
Once this migration has been run `studies/go_viral` can be removed completely (along with the contents of this migration, which is necessary because the test suite will fail on this migration after `studies/go_viral` is removed).


private_sharing/migrations/0003_auto_20160909_0427.py
```diff
@@ -0,0 +1,99 @@
...
+        while DataRequestProjectMember.objects.filter(
+                project_member_id=code).count() > 0:
+            code = generate_id(size=8, chars=digits)
+
+        return code
+
+    project = OnSiteDataRequestProject(
+        is_study=True,
+        name='GoViral (2014-2016)',
+        slug='go-viral-archive',
```
madprime: I thought the slugs are automatically created based on the name?

beaugunderson: they are, I meant to remove this because it has no effect, good catch :)

private_sharing/migrations/0003_auto_20160909_0427.py
```diff
@@ -0,0 +1,99 @@
...
+        leader='Dr. Rumi Chunara',
+        organization='NYU Polytechnic School of Engineering',
+        is_academic_or_nonprofit=True,
+        contact_email='rumi.chunara@nyu.edu',
```
madprime: Should change to current contact email GoViral is using for its reply-to: goviralstudy@gmail.com

private_sharing/migrations/0003_auto_20160909_0427.py
```diff
@@ -0,0 +1,99 @@
...
+        info_url='https://www.goviralstudy.com/',
+        short_description="""Participants in this viral surveillance study can
```
madprime: Short: Participants in this viral surveillance study received kits, and sent samples if they got sick.

Long: This project represents previous years of the GoViral study. Participants in this viral surveillance study received kits, and sent samples if they got sick. Data import from this study is no longer available, but data remains in member accounts if it was previously imported.

(Of course, we can change these later via admin panel too...)

private_sharing/migrations/0003_auto_20160909_0427.py
```diff
@@ -0,0 +1,99 @@
...
+    def random_id():
+        code = generate_id(size=8, chars=digits)
+
+        while DataRequestProjectMember.objects.filter(
+                project_member_id=code).count() > 0:
+            code = generate_id(size=8, chars=digits)
+
+        return code
+
+    project = OnSiteDataRequestProject(
```
madprime: I think an OAuth2 project would make more sense -- technically that's what it was (albeit custom). One advantage is that this doesn't include the "consent" form, which doesn't make sense in this context.

madprime:
Minor point, but also need to copy public data sharing status over to the new project (and maybe delete the old info).

madprime:
(Other than that, looks good to me! :+1: )

PR #591, opened 2016-09-25
S3 direct uploads

beaugunderson:
- implements S3 direct upload
- refactors some form handling logic to be much cleaner
- adds comments
- adds tests for file deletion


open_humans/management/commands/expunge_project_data_files.py
```diff
@@ -0,0 +1,28 @@
...
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+from private_sharing.models import ProjectDataFile
+
+
+class Command(BaseCommand):
+    """
+    A management command for expunging incomplete project data files.
```
madprime: At the risk of creating redundant and/or stale docstrings, I'd like to see a bit more context provided. Like what makes a data file incomplete? When are things expunged?

In this case, maybe something like this?

ProjectDataFile objects which are not complete (complete=False) represent incomplete upload attempts. (See ViewReference1 and ViewReference2.) This command deletes objects that have been incomplete for longer than the period specified by settings.INCOMPLETE_DATA_FILE_EXPIRATION.

private_sharing/api_views.py
```diff
@@ -130,103 +124,154 @@ def post(self, request, format=None):
...
+        }, status=status.HTTP_201_CREATED)
 
-        form = DeleteDataFileForm(request.data)
 
-        if not form.is_valid():
-            return Response({'errors': form.errors},
-                            status=status.HTTP_400_BAD_REQUEST)
+class ProjectFileDirectUploadCompletionView(ProjectFormBaseView):
+    """
+    Initiate a direct upload to S3 for a project by pre-signing and returning
```
madprime: Copy pasta :)

private_sharing/api_views.py
```diff
@@ -130,103 +124,154 @@ def post(self, request, format=None):
...
-        file_id = form.cleaned_data['file_id']
-        file_basename = form.cleaned_data['file_basename']
-        all_files = form.cleaned_data['all_files']
+    def post(self, request):
+        super(ProjectFileDirectUploadCompletionView, self).post(request)
+
+
+class ProjectFileUploadView(ProjectFormBaseView):
+    """
+    A form for regular ProjectDataFile uploads.
```
madprime: What's a "regular" ProjectDataFile upload? (I'm pretty sure I know what you mean -- but let's make some docstrings that give a bit more context for future code navigators. :) )

PR #598, opened 2016-09-30
Cache activities for logged out users

beaugunderson:


beaugunderson:
merging since @madprime gave the 👍 via Slack

PR #648, opened 2017-03-28
Add Github to footer as we are an opensrouce project

BenjaminHCCarr:
It is very hard to find the code, and since we are FOSS/OpenSource, I think adding the Github badge with a link to the main organization page made sense.

The only mention is in about, and then still buried.


BenjaminHCCarr:
`stash`

PR #649, opened 2017-03-28
Move 'Forums' from Footer to Header

BenjaminHCCarr:
To try to increase interaction in the forums, I am submitting a PR to move the link from the footer to the header. With thousands of OH users and only a couple dozen forum members, I am not sure people know it's available. Moving it from buried in the footer to the header will hopefully drive up participation.

Users know to look for
- twitter
- facebook
- github
- etc

in the footer, but our forums are unique and underutilized.


BenjaminHCCarr:
`stash`

PR #675, opened 2017-11-22
fix bug on member list and public data pages

beaugunderson:
as best as I can tell something in either jQuery or select2 changed in the last year so that when `require()`ing select2 one must pass in the jQuery object as well.

the unrelated dependency updates are what was required to get the app to run in my environment so that I could test the fix. :)


gedankenstuecke:
LGTM, but as I have no clue about the deployment yet I'll refrain from merging it 😂

PR #684, opened 2017-12-22
Add joined projs & connected data to account pages

madprime:
This is in preparation with a redesign which separates out
the activities lists (currently on the front page) into
separate "data source" and "sharing" pages. This will
result in the "Connected & Joined" option disappearing;
these new pages provide a replacement for users that seeking
that information.


open_humans/templates/member/my-member-connected-data.html
```diff
@@ -60,26 +66,14 @@
         {% endfor %} {# data_file in data_files #}
       </table>
     {% else %}
-      <p>There are no files associated with this source.</p>
+      <p>There are not currently any files associated with this source.</p>
```
gedankenstuecke: Super weird phrasing. Should probably be `There are currently no files associated with this source`.

open_humans/templates/member/my-member-connected-data.html
```diff
@@ -4,30 +4,36 @@
...
-  <p>Below is a summary of your data files in Open Humans.</p>
+  <h2>Connected data</h2>
+
   <p>
-    You can click on a particular data source to get options for managing your
-    incoming data and sharing it with projects.
+    You have connected to the following activities that add data to your
+    account. Click on a project to visit its activity page and learn
+    more about it, deauthorize it, or send a message to project
+    coordinators.
```
gedankenstuecke: Maybe it's already taken care of, but I couldn't figure it out by reading the code: What happens if I disconnect from a data source but the files remain in my account. Will it be listed in my data sources here? If yes, the description here is a bit misleading.

madprime: Oh... yes, you're right, they will also be listed here. Anything that is classified as a "data source" and/or has files will show up.

gedankenstuecke: In that case it would probably be good to split the view into `active projects` on the top and `data from disconnected projects` on the bottom?

madprime: Ah, yes, that makes sense, will do!

PR #687, opened 2017-12-28
Remove redundant JS imports

### LINKED ISSUE #686: Navbar menu won't collapse on small/mobile views
The navbar menu opens on a click in small/mobile view – but clicking again does _not_ close it as it should.

This bug seems to be caused by a double inclusion of jQuery – per [this response on Stack Overflow](https://stackoverflow.com/a/26403469)? jQuery is first added in `main.js` ([see here](https://github.com/OpenHumans/open-humans/blob/28c9a6fb9ad94f666a74ec658824c6f8b25aaa35/static/js/main.js#L6)) and often (but not always) another time in a page-specific JS, e.g. `home.js` ([see here](https://github.com/OpenHumans/open-humans/blob/28c9a6fb9ad94f666a74ec658824c6f8b25aaa35/static/js/home.js#L3)).

It's unclear how long this issue has been on the site… if this diagnosis is correct, this double-includes might be causing other issues.

madprime:
Fixes #686.

### DIFF
--- a/static/js/account-password-reset-token.js
+++ b/static/js/account-password-reset-token.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var $ = window.jQuery = require('jquery');
+var $ = window.jQuery;
 
 $(function () {
   $('#id_password').attr('required', '');
--- a/static/js/account-password.js
+++ b/static/js/account-password.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var $ = window.jQuery = require('jquery');
+var $ = window.jQuery;
 
 $(function () {
   $('#id_password_current').attr('required', '');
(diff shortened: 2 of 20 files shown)


madprime:
@gedankenstuecke I think this fixes the navbar menu collapse issue… but I'm unsure about myself here with this JS stuff! I _think_ the deal is that netlify was importing multiple times with each `require`.

I wonder if maybe @beaugunderson has a comment on whether these were indeed leading to redundant imports? (Your feedback is always really really appreciated! But between holidays & your birthday I appreciate this might not be the best time to pester you. :) )

gedankenstuecke:
Seems like these were not really redundant imports after all: 

screen shot 2017-12-28 at 13 24 41

The member search doesn't work any longer, so somewhere things have gone wrong. Good news though: yes, the menu works now ;)

PR #697, opened 2018-01-03
allow users to delete their 3rd party project files when disconnecting.

### LINKED ISSUE #685: Clarify that deauthorizing projects doesn't delete data
An activity's page gives the information that `You can withdraw and deauthorize this project at any time to immediately revoke all granted permissions. Click here to withdraw/deauthorize this project.` It does not make clear that the files it put into your account will not be deleted. This leads to an unclear behavior for the user. 

You might expect that files will automatically be deleted as well which could lead to you 
* either not leaving a project you want to leave because you want to keep the data (which is what happens already, but you don't know this)
* or you are leaving a project, expected that your (potentially public) data will be gone as well (which is not what happens, leading to accidental data leakage for that user). 

To solve this I propose an extra page that appears after a user clicks on an updated `Click here to withdraw/deauthorize this project.` button. This new page would offer the two options of `Leave project X and keep the data it deposited in your account` and `Leave project X and delete all data it deposited`. This way both issues mentioned above would be solved. 

In a simpler case these two options could also just be put on the bottom of the activity page, where the single link is already placed right now.

gedankenstuecke:
Would close #685 by allowing the optional deletion of data files attached to a project when disconnecting the project.

### DIFF
--- a/private_sharing/templates/private_sharing/leave-project.html
+++ b/private_sharing/templates/private_sharing/leave-project.html
@@ -13,11 +13,19 @@
 </div>
 
 <div class="row">
-  <div class="col-xs-12 text-right">
+  <div class="col-xs-12">
     <form method="POST" action="">
       {% csrf_token %}
 
-      <p>
+      <div class="checkbox">
+        <label>
+          <input type="checkbox" name="remove_datafiles">
+
+          Also remove my data files for {{ object.project.name }}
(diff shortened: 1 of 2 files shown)


madprime:
Works for me! I tested locally: created a project, joined, added a file, and tried both the don't-delete and delete conditions.

PR #700, opened 2018-01-10
remove usersnap applet

### LINKED ISSUE #168: Disable usersnap for mobile visitors
It's not obvious what is is and I don't think there's a way to customize it.

gedankenstuecke:
Removes the Usersnap applet as discussed on Slack and closes #168.

(also removes the picture of Liz which is not used anymore)


madprime:
This seems to remove Liz Salmi's photo?

gedankenstuecke:
whops, my bad. will fix

Issue #699, opened 2018-01-10
Statistics


gedankenstuecke:
@madprime and I discussed that we should have some page that lists statistics about Open Humans that can give some sense of how large & active our community is (especially as our GH pages do not do justice to all the coding that individual people put into their own projects). 

A statistics page should include details on:

* Number of approved & running projects
* Maybe the number of closed projects -> sense of how many studies have already been done?
* Number of projects that are in development right now
* Number of users that joined Open Humans
* Maybe the number of total connections of member <-> projects that were created so far?

Feel free to leave comments with interesting statistics that capture community engagement that I missed so far!

madprime:
Closed by #708

gedankenstuecke:
Whops, thanks for closing it!

PR #708, opened 2018-01-17
add statistics draft

gedankenstuecke:
For #699

### DIFF
--- a/open_humans/views.py
+++ b/open_humans/views.py
@@ -14,6 +14,7 @@ from django.http import Http404, HttpResponseRedirect
 from django.shortcuts import render
 from django.views.generic.base import TemplateView, View
 from django.views.generic.edit import DeleteView, FormView
+from django.db.models import Count
 
 import feedparser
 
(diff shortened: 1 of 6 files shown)


open_humans/views.py
```diff
@@ -516,6 +517,63 @@ def form_valid(self, form):
...
+    def get_number_member():
+        members = Member.objects.all()
+        members_with_data = members.annotate(
+            datafiles_count=Count('user__datafiles')).filter(
+            datafiles_count__gte=1)
+        return (len(members), len(members_with_data))
+
+    @staticmethod
+    def get_number_files():
+        files = DataFile.objects.all()
```
beaugunderson: `return DataFile.objects.count()` :)

open_humans/views.py
```diff
@@ -516,6 +517,63 @@ def form_valid(self, form):
...
+class StatisticView(NeverCacheMixin, SourcesContextMixin, TemplateView):
+    """
+    Show latest statistics on signed up users/projects etc.
+    """
+
+    template_name = 'pages/statistics.html'
+
+    @staticmethod
+    def get_number_member():
+        members = Member.objects.all()
```
beaugunderson: `Member.objects.count()`

open_humans/views.py
```diff
@@ -516,6 +517,63 @@ def form_valid(self, form):
...
+        members_with_data = members.annotate(
```
beaugunderson: I think this is equivalent if you just need any member with > 0 data files: 

```python
members_with_data = members.filter(user__datafiles__isnull=False).count()
```

gedankenstuecke: I followed @madprime's lead here from https://github.com/OpenHumans/open-humans/blob/master/open_humans/management/commands/stats_new.py 🙂

open_humans/views.py
gedankenstuecke: 👍

open_humans/views.py
madprime: I probably didn't know about (or remember) `isnull` :)

PR #711, opened 2018-01-26
Add filters on is_active User status

madprime:
...thereby enabling account suspensions.

### DIFF
--- a/common/activities.py
+++ b/common/activities.py
@@ -69,7 +69,7 @@ def badge_counts_inner():
     Return a dictionary of badges in the form {label: count}; e.g.
     {'fitbit': 100}.
     """
-    members = Member.objects.all().values('badges')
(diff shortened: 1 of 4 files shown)


madprime:
@gedankenstuecke No, it returns an authentication error, as if they have the wrong password. And a request for a password reset email seems to fail silently. (No error reported, no email sent?) I think this is OK but could definitely be more user-friendly.

gedankenstuecke:
Okay, sounds good to me. Maybe should add to the suspension email that they need to email us as well if they want us to fully delete the account then.

madprime:
I'll merge & let you create new issues or edit email as you see fit :)

gedankenstuecke:
Ok, just added to the suspension email to make the consequences more clear.

PR #713, opened 2018-01-26
For messaging, check member hasn't quit project

### LINKED ISSUE #712: Project messages received after deauthorization
A member reports deauthorizing a project, but then receiving a message from them – which shouldn't be possible. It sounds like a bug.

madprime:
Fixes #712

### DIFF
--- a/private_sharing/forms.py
+++ b/private_sharing/forms.py
@@ -142,9 +142,10 @@ class MessageProjectMembersForm(forms.Form):
                 'Project member IDs are always 8 digits long.')
 
         # look up each ID in the database
-        project_members = DataRequestProjectMember.objects.filter(
-            project_member_id__in=project_member_ids,
-            message_permission=True)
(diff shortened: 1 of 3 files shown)


private_sharing/forms.py
```diff
@@ -142,9 +142,10 @@ def clean_project_member_ids(self):
                 'Project member IDs are always 8 digits long.')
 
         # look up each ID in the database
-        project_members = DataRequestProjectMember.objects.filter(
-            project_member_id__in=project_member_ids,
-            message_permission=True)
+        project_members = (DataRequestProjectMember.objects
```
gedankenstuecke: Does `.filter_active()` remove all the ones that have revoked access?

madprime: yup see https://github.com/OpenHumans/open-humans/blob/4998871fc51d6a4466030bf031a3f5f8a53b807a/private_sharing/models.py#L309

Issue #724, opened 2018-02-13
Link to API for the connected member list from activity page


madprime:
The Public Data API has an endpoint for querying members by source. In activity pages, the section that states "Connected by N members" could be linking to this API.

An additional issue to address at the same time: this link should be using the API with a filter-by-source, which doesn't seem to be a feature it currently has.

delenamalan:
I'm keen to work on this.

madprime:
Hi @delenamalan! This is an issue in our main site – which is open source, but complex (it only runs in one place) and getting to the point that one can contribute to that takes quite a while. I can do this task pretty easily myself.

The best opportunities for newcomers to contribute will be with new projects (that work with this main site's APIs). :) Sometimes those needs get documented on the main site, but become modular projects in new repos – e.g. https://github.com/OpenHumans/open-humans/issues/690

(This is true for our staff hires as well. The main site is a tough place to start!)

We're working on fleshing out some documentation and suggestions for more tractable opportunities for contribution – I'm sorry it's not already available, please stay tuned in the Slack channel!

delenamalan:
@madprime no problem. I only saw your message now so I worked on it anyway. If you don't need it, it was a fun exercise in any case. 😸

madprime:
@delenamalan oh gosh, wow! And I was so worried about someone getting lost in our main site. :D You rock. I might not have time to review this until tonight – thank you!

gedankenstuecke:
I assume this can be closed thanks to #725?

madprime:
Yup! closed by #725 – thank you @delenamalan!

PR #725, opened 2018-02-13
Link to API for the connected member list from activity page

delenamalan:
Addresses #724

### DIFF
--- a/open_humans/api_views.py
+++ b/open_humans/api_views.py
@@ -10,7 +10,6 @@ from rest_framework.generics import ListAPIView
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
-from common.utils import app_label_to_verbose_name
 from data_import.models import DataFile
 from private_sharing.utilities import (
     get_source_labels_and_names_including_dynamic)
@@ -127,9 +126,6 @@ class PublicDataUsersBySourceAPIView(APIView):
                 sources[badge['label']].append(user['username'])
 
         source_filter = request.GET.get('source')
-        source_labels_and_names = [
-            (source_filter, app_label_to_verbose_name(source_filter)),
-        ] if source_filter else get_source_labels_and_names_including_dynamic()
 
         source_list = [
             {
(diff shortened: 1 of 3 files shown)


BenjaminHCCarr:
The code looks sane/non-malicious, not sure if it's feature complete unless the import is calling something I haven't seen before.

open_humans/api_views.py
```diff
@@ -125,14 +126,19 @@ def get(request):
 
                 sources[badge['label']].append(user['username'])
 
+        source_filter = request.GET.get('source')
+        source_labels_and_names = [
+            (source_filter, app_label_to_verbose_name(source_filter)),
```
madprime: sorry for the complexity in the backend here! some of the sources are "projects" and others are "apps" – this function only works for apps. (fwiw "apps" is deprecated :sweat:)

(I have a rewrite ready that avoids this and another issue – just leaving notes :) )

open_humans/templates/partials/activity-info-table.html
```diff
@@ -50,8 +50,10 @@
     <td>
       <ul>
         <li>
+          <a href="{% url 'members-by-source' %}?source={{ activity.source_name }}">
```
madprime: It turns out this `get_source_labels_and_names` only includes projects that are acting as data sources... which, to be fair, seems consistent with the API's description.

Maybe it would be nice to change the API's behavior on this, but keeping that the same, we should only render this link on projects when it's judged to be a "data source", consistent with the API – which is based on whether the `returned_data_description` is blank: https://github.com/OpenHumans/open-humans/blob/master/private_sharing/utilities.py#L17

open_humans/api_views.py
```diff
@@ -133,6 +135,8 @@ def get(request):
             }
             for label, verbose_name
             in get_source_labels_and_names_including_dynamic()
+            if (not source_filter) or
+            (source_filter and label in source_filter.split(','))
```
delenamalan: 🥇

PR #730, opened 2018-02-15
Fixed the target link.

### LINKED ISSUE #723: My account > Public profile should be the public profile
It should go to `/user/username` – not to the dashboard at `/member/me`.

And maybe the latter is entirely unnecessary. There could be an "edit profile" a user sees on the profile page itself, if a user is viewing their own profile.

troublemagnet:
Closes #723

### DIFF
--- a/open_humans/templates/base.html
+++ b/open_humans/templates/base.html
@@ -94,7 +94,7 @@
                   aria-haspopup="true" aria-expanded="false">My account
                   <span class="caret"></span></a>
               <ul class="dropdown-menu">
-                <li><a href="{% url 'my-member-dashboard' %}">Public profile</a></li>
+                <li><a href="{% url 'member-detail' member.user.username %}">Public profile</a></li>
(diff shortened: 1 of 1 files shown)


open_humans/templates/base.html
```diff
@@ -94,7 +94,7 @@
                   aria-haspopup="true" aria-expanded="false">My account
                   <span class="caret"></span></a>
               <ul class="dropdown-menu">
-                <li><a href="{% url 'my-member-dashboard' %}">Public profile</a></li>
+                <li><a href="{% url 'member-detail' member.user.username %}">Public profile</a></li>
```
madprime: Can we change this to 'member-detail-direct' instead?

'member-detail' is nice to retain as an unambiguous fallback, but I think people like having a profile URL that looks like openhumans.org/username :)

gedankenstuecke:
Closing this for now as it hasn't been updated. Please feel free to make a new PR that addresses the feedback of @madprime.

PR #732, opened 2018-02-16
templates for issue/PR submission

### LINKED ISSUE #729: Need for "Create a PR" and "Create an issue" templates
For standardizing the pull requests body and for making it easy to find relevant issues, there should be respective templates IMO. 
By having a PR template with a Closes/Fixes entry, it can be ensured that each of the PRs map to some issue. Also making it mandatory to have some description of what has been changed/implemented in the PR, will make it convenient for maintainers for review.
Apart from that an issue template, will make it easier for finding relevant issue. While this isn't as important as "Creating a PR" template (You can always label the issues to indicate the type,instead of putting it in the title), but could be a great addition to standardize issues.

gedankenstuecke:
I expanded on @BenjaminHCCarr's suggestion (closes #731 and closes #729) by riffing of the templates of `ropensci/taxize`. It makes more use of HTML comments so that people see what is expected in the submissions.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,21 @@
+<!--- Provide a general description of your issue in the Title above -->
+
+## General Checkups
+- [ ] Have you checked that there isn't already an existing issue that describes what you report below?
+- [ ] Have you checked that there isn't already an open pull requests for this issue/update/change?
+
+<!---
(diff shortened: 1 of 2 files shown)


madprime:
Just to check, do we know this works when the template filename is lowercase? Since it's uppercase filenames in https://github.com/blog/2111-issue-and-pull-request-templates

gedankenstuecke:
Yes, that works, `taxize` which I used as a template for the comments has this as well and there it works (c.f. https://github.com/ropensci/taxize/tree/master/.github)

PR #734, opened 2018-02-23
Fix number of members

### LINKED ISSUE #733: Statistics page should exclude inactive accounts
## Description
The statistics page `/statistics/` displays an incorrectly large number of members. This is probably due to a large number of "spam" accounts that were rendered inactive.

The fix is probably to to add a filter `user__is_active=True` to the Member query in StatisticView in `open_humans.views`.

## Related Issue(s)
NA

## Example
Currently [`/statistics/`](https://www.openhumans.org/statistics/) page says "5864 people have joined us", but the [`/members/`](https://www.openhumans.org/members/) page lists 5,075.

gedankenstuecke:
## General Checkups

## Related Issue

Closes #733

### DIFF
--- a/open_humans/views.py
+++ b/open_humans/views.py
@@ -526,7 +526,7 @@ class StatisticView(NeverCacheMixin, SourcesContextMixin, TemplateView):
 
     @staticmethod
     def get_number_member():
-        members = Member.objects.all()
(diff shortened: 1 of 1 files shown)


gedankenstuecke:
If the tests pass I'll merge to staging so we can see whether it gives the right answer 😉

PR #736, opened 2018-02-25
Fix URL error for projects with underscore in name

### LINKED ISSUE #735: Activity pages have errors, won't load!
## Description
Activity pages won't load! The error seems to be specific to production, potentially related to a project created on the main site.

madprime:
## Description
Fixes #735!

This is an urgent issue on the site.

Once the fix is merged to master, I'll update production to match master.


madprime:
Good enough – need to get this fixed, I'll merge :)

Issue #739, opened 2018-03-14
Add a page listing project grants


gedankenstuecke:
## General Checkups
- [x] Have you checked that there isn't already an existing issue that describes what you report below?
- [x] Have you checked that there isn't already an open pull requests for this issue/update/change?

## Description
We need a page that lists the current & past project grants, with links to the repos, a description, the project grant awardee etc. @mcescalante is already on it, but I wanted to have this down here too. 😄

mcescalante:
All that's left here is for me to rebase against master to get updates since I worked on this last and open a PR

PR #787, opened 2018-05-16
Add grant projects page

mcescalante:
## General Checkups

## Description

This PR adds a "Grant Projects" page to the website as requested by #739, and the related models, views, and templates for those grant projects. It also includes a fixture to seed data initially to the tables, which may become outdated as we add and adjust projects, but it will be useful for the initial import.

Right now, there is no link added to the main site in this PR to point users to this new Grant Projects page, so that is something that should be decided and added in before this is merged, most likely. I also did not include photos yet.

### DIFF
--- a/open_humans/templates/pages/project-grants.html
+++ b/open_humans/templates/pages/project-grants.html
@@ -40,6 +40,10 @@
   <li>Connect to an existing community or study</li>
 </ul>
 
+<p>
+  Check out previous awardees on our <a href="{% url 'grant-projects' %}">project grant awards page</a>!
+</p>
+
 <iframe class="visible-xs-block" frameborder="0" width="448" height="252"
   style="margin:0 auto 10px;"
   src="https://www.youtube.com/embed/q97lIpuL-bs"


package.json
```diff
@@ -10,8 +10,8 @@
     "url": "https://github.com/OpenHumans/open-humans.git"
   },
   "engines": {
-    "node": "6.x",
-    "npm": "3.x"
+    "node": "6.x || 7.x || 8.x || 9.x",
```
gedankenstuecke: Can we keep the versions as is? I guess at least we don't need these bumps for the project grants page and by keeping this constant we have a better atomic commit.

open_humans/models.py
```diff
@@ -257,3 +257,24 @@ def create(cls, rss_feed_entry):
...
+    """
+    name = models.CharField(max_length=255, unique=True)
+    grant_date = models.DateField(null=True)
+    status = models.CharField(max_length=120)
+    github = models.TextField(blank=True)
+    grantee_name = models.CharField(max_length=255)
+    photo = models.ImageField(
+        blank=True, 
+        max_length=1024,
+        # Stored on S3 TODO: update these fields
```
gedankenstuecke: Is this still a todo? I've no idea how the S3 uploads currently work for other photos.

open_humans/fixtures/grant-projects.json
```diff
@@ -0,0 +1,58 @@
...
+      "grant_date": "2018-02-20",
+      "status": "Approved",
+      "github": "",
+      "grantee_name": "Kevin Arvai",
+      "photo": "",
+      "blog_url": "https://blog.openhumans.org/2018/02/20/interviewing-project-grant-awardee-kevin-arvai/",
+      "project_desc": "The goal of Imputer is to provide users with a more comprehensive picture of their genome. Direct to consumer genetics companies, like 23andMe, only genotype a small fraction of the genome. Researchers are finding new genetic locations associated with traits and diseases at a rapid pace. Users might be interested in knowing their genotype status for these new associations, but the locations may be in regions that direct to consumer tests are not genotyping. Imputer leverages the vast amount of genotype data made available by 1000 genomes project and by the Haplotype Research Consortium to provide Open Humans users with genotype estimates at additional locations in their genome."
+    }
+  }
+]
```
gedankenstuecke: This is missing the project grant for andrew riha: https://blog.openhumans.org/2018/03/23/meet-andrew-riha-our-next-project-grant-awardee/

madprime: I think we shouldn't have any fixture at all - my understanding is we want to use the Django admin interface to add and edit these?

mcescalante: You can use the django admin interface to edit them (add, edit, remove, etc.), the fixture is just there to assist seeding data for the 1st run / testing locally 😄

open_humans/models.py
mcescalante: I copied the image storage method from other models and left that TODO for myself. I believe this will work properly, but I am not sure how the uploads work either so maybe that would be good to understand before I make changes (or not) to the photo bit

open_humans/migrations/0010_grantproject_grant_date.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.9 on 2018-03-14 21:32
+from __future__ import unicode_literals
```
madprime: Since this hasn't hit the production database yet, I'd prefer to have just one migration in the final merge (i.e. delete the two migrations and run `makemigrations` again).

open_humans/admin.py
```diff
@@ -1,7 +1,12 @@
...
 from django.contrib.auth.admin import UserAdmin
 
-from .models import Member, User
+from .models import Member, User, GrantProject
 
 admin.site.register(Member)
 admin.site.register(User, UserAdmin)
+
+class GrantProjectAdmin(admin.ModelAdmin):
+    list_display = [field.name for field in GrantProject._meta.get_fields()]
```
madprime: Out of curiosity, how is this different than the default Django model admin?

open_humans/fixtures/grant-projects.json
madprime: I think we should skip adding the fixture, since it's a one-off for initial data, and we don't really need it. :)

open_humans/admin.py
mcescalante: There only difference is that this allows us to display the details in a list in the admin interface without clicking through to each item:

With this ModelAdmin addition (`id` clicks through):


Without (the name clicks through):


Happy to remove it if we prefer to leave the admin vanilla/default!

madprime: I like the improved version, please keep it!

open_humans/models.py
madprime: I'm comfortable with treating it as if it works, and if it doesn't, fixing it later. But maybe removing the TODO? :)

open_humans/urls.py
```diff
@@ -110,6 +110,9 @@
     url(r'^public-data-api/$',
         views.PublicDataDocumentationView.as_view(),
         name='public-data-api'),
+    url(r'^grant-projects/$',
+        views.GrantProjectView.as_view(),
+        name='project-grants'),
```
madprime: I think we'd best rename the name to `'grant-projects'` to match the template and view and URL, especially since there's already a template `project-grants.html`

PR #797, opened 2018-06-18
Remove legacy apps - stage 1

madprime:
## Description
This is stage 1 of app removal, which needs to be a staged process.

In this first stage, the models have been removed and migrations are created that will be run to remove the models from the database. Because migration files reside _within_ the apps, removal of the apps themselves should only occur after running the migrations.

In addition to migrations to remove obsolete models, this PR removes obsolete code related to working with the obsolete "data processing" Flask server.

@beaugunderson – I'm flagging you as a reviewer along with @gedankenstuecke in case you have some comments on this plan. No worries if you don't have time, but your thoughts would be very welcome. (via Slack is good too) :)


data_import/models.py
```diff
@@ -6,14 +6,10 @@
 from django.core.urlresolvers import reverse
 from django.db import models
```
gedankenstuecke: Do we need these models at all after the removal? or is that because of the weird "project data being a subclass" thing that we need to keep it around for now?

gedankenstuecke:
Btw. I just added `**/migrations/` to the `codeclimate` ignore-list, as it seems pointless for it to fail for code duplication on stuff that's automatically generated by the framework.

data_import/models.py
madprime: To respond here - the DataFile model is still used, but as an obsolete non-abstract parent to a child class (which is used for everything now)... and there are no other children. So this could definitely be cleaned up, but since it's nontrivial, not worrying about it here. :)

madprime:
Thanks @gedankenstuecke! And thanks for updating codeclimate with that

PR #805, opened 2018-07-10
add url_slug for public data

### LINKED ISSUE #798: Public data activity is broken
## General Checkups

## Description
the public sharing feature seems to be broken right now, e.g. if you go to https://www.openhumans.org/member/me/joined/ and are joined in the public data sharing study the link leads to https://www.openhumans.org/activity/None/

BenjaminHCCarr:
This had been returning "NONE" the fallthrough in the python script. I believe this should close #798

You might want to test this on staging, but given that it fails in current state...

This was based on the `url_slug` in the `apps.py` of ancestry_dna page.

## General Checkups
- [X] Have you checked that there aren't other open pull requests for the same issue/update/change?

## Related Issue
fix #798

### DIFF
--- a/common/activities.py
+++ b/common/activities.py
@@ -240,7 +240,7 @@ def public_data_activity(user):
         'type': 'internal',
         'connect_verb': 'join',
         'join_url': reverse('public-data:home'),
-        'url_slug': None,
+        'url_slug': reverse('public-data:home'),
         'is_connected': (user and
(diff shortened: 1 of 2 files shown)


madprime:
Looks like @gedankenstuecke has excavated the correct location the code should be changed! Kind of gnarly but yeah, that's where it's happening. :grimacing:

PR #807, opened 2018-07-11
prevent activity stream duplicates

### LINKED ISSUE #801: Activity stream updates for public sharing
## General Checkups

## Description

If you stop publicly sharing a data source this also triggers a `user XXX publicly shared data source YYY` event on the frontpage. E.g. I started & stopped sharing my Twarxiv data and it yielded two separate events on the front page.

gedankenstuecke:
## General Checkups

## Description

The activity stream would show duplicate entries for re-joining a project a member was already a member of but where the external project would require new oauth credentials and similarly when data was being made public a 'second time'. These could look weird on the frontend. 

This PR would fix these by not adding a second item to the stream if people have already done the action once.

## Related Issue

Closes #801

### DIFF
--- a/private_sharing/views.py
+++ b/private_sharing/views.py
@@ -91,7 +91,10 @@ class ProjectMemberMixin(object):
         django_messages.success(self.request, (
             'You have successfully joined the project "{}".'.format(
                 project.name)))
-        if project.approved:
+        if project.approved and not ActivityFeed.objects.filter(
+           member=self.project_member.member,
+           project=project,
+           action='joined-project').exists():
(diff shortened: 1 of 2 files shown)


gedankenstuecke:
I've tested the code on my local dev DB, but only for the `joined-project` logic, not the `public-shared`, as I don't have the public sharing thing set up right now. But that should be identical.

gedankenstuecke:
That's cool by me, actually the linter didn't complain about overindent, so I did that now :)

gedankenstuecke:
Thanks!

PR #815, opened 2018-07-18
Make GitHub recognize the LICENSE

abitrolly:
GitHub should display MIT badge in top right corner if license is recognized.

image

Current page shows this.

image

Even if this PR doesn't change license text much, it is still able to kick the GitHub license match job.

As a benefit, this project could be then filtered by MIT License.


abitrolly:
image
All good now.

PR #817, opened 2018-07-18
Migrate to Python3

jhdulaney:
## General Checkups

## Description
Migrated site to python 3.  Currently passes all tests with one exception, which is due to a bug in upstream django (#27398).  Also manually tested the site.


open_humans/management/commands/suspend_users.py
```diff
@@ -47,4 +47,4 @@ def handle(self, *args, **options):
         for user in users_to_suspend:
             user.is_active = False
             user.save()
-            print('{} (ID: {}) is suspended.'.format(user.username, user.id))
+            print('{} (ID: {}) is suspended.'.format(user.username, user.id)
```
madprime: I don't think management commands are particularly important, but this looks like it might be a typo?

jhdulaney: Clearly that does not have a test.  It was for sure a typo, and I'll be pushing a fix momentarily.

jhdulaney:
Fixed the typo, updated requirements.txt, will be pushing changes momentarily.

jhdulaney:
And, changes pushed.

PR #832, opened 2018-08-16
Enable projects to remove project members

### LINKED ISSUE #737: Stats of project is inaccurate; enable project admins to remove non-participants (Technical and Organizational issue)
**Issue:** the "stats" of a project reflects the number of people who click "join" on a project, even if they're not legitimate/valid participants (e.g. don't have required data sources). 

Related to the project admin tools that need to be built (https://github.com/OpenHumans/open-humans/issues/690), we need to decide:
- If project admins can 'remove' participants
- Do 'removed' participants get notified? Should they?

#### Example

screen shot 2018-03-07 at 3 06 48 pm

People cite this as "95 OpenAPS users have donated data". This is incorrect. This number is inclusive 50-75% which are OH accounts that have "joined" the project despite not having OpenAPS data, and I would like a way to remove non-valid data sets and associated accounts so this number is an accurate reflection of this data commons.

madprime:
## General Checkups

## Description
Add on-site and API endpoint to allow projects to remove members from the project.

## Remaining TODOs

- [x] Add tests
- [x] Add documentation
- [x] Reduce redundant code redundancy

## Related Issue
Closes #737

### DIFF
--- a/private_sharing/templates/private_sharing/remove-project-members.html
+++ b/private_sharing/templates/private_sharing/remove-project-members.html
@@ -1,0 +1,30 @@
+{% extends 'panel.html' %}
+
+{% load bootstrap_tags %}
+{% load static from staticfiles %}
+
+{% block head_title %}{{ aoeueoa }}Remove project members{% endblock %}
+
+{% block panel_content %}
+
+<h4>Project: {{object.name}}</h4>
+
+<p>
+  <b>CAUTION: this action cannot be reversed by you!</b><br>
+  You can use the form below to remove members from your project.
+</p>
+
+<div class="row pad-all-sides">
+  <form class="form-vertical" role="form" method="post" rel="persist"
+    action="." id="remove-members-form">
+
+    {% csrf_token %}
+
+    {{ form|as_bootstrap }}
+
+    <input id="remove-members" type="submit"
+      value="Remove members"
+      form="remove-members-form" class="btn btn-danger">
+  </form>
+</div>
+{% endblock %}
--- a/private_sharing/tests.py
+++ b/private_sharing/tests.py
@@ -488,6 +488,11 @@ class SmokeTests(SmokeTestCase):
         '/direct-sharing/projects/on-site/create/',
         '/direct-sharing/projects/on-site/update/abc-2/',
         '/direct-sharing/projects/on-site/abc-2/',
+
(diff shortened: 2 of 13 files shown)


private_sharing/tests.py
```diff
@@ -458,6 +488,11 @@ class SmokeTests(SmokeTestCase):
         '/direct-sharing/projects/on-site/create/',
         '/direct-sharing/projects/on-site/update/abc-2/',
         '/direct-sharing/projects/on-site/abc-2/',
+
+        # Tests desired appear to fail due to missing 'css_classes', probably
```
jhdulaney: https://docs.djangoproject.com/en/2.1/ref/forms/api/#django.forms.BoundField.css_classes

Seems to be still supported?

madprime: Maybe how it's done shifted? I really don't know. I couldn't easily see how to make the test not fail, unfortunately, but I left this here in the hopes that someone else might solve it. :-/

private_sharing/templates/private_sharing/remove-project-members.html
```diff
@@ -0,0 +1,30 @@
...
+{% load static from staticfiles %}
+
+{% block head_title %}{{ aoeueoa }}Remove project members{% endblock %}
+
+{% block panel_content %}
+
+<h4>Project: {{object.name}}</h4>
+
+<p>
+  <b>CAUTION: this action cannot be reversed by you!</b><br>
```
jhdulaney: Besides maybe going into the db and manually re-adding, I don't quite see an easy way to reverse this?  Maybe have something in the admin page?

private_sharing/tests.py
jhdulaney: I'll add it to my todo list under the tests section ...

private_sharing/templates/private_sharing/remove-project-members.html
madprime: Another way to reverse it is that the individual can re-join... I hope projects don't screw up in removals, but if they do, yeah I guess they could petition us to do something like you describe. I don't envision it coming up (and hopefully never), so wouldn't worry about making an easy tool for doing it...

madprime:
I went and marked some of the codeclimate issues as invalid – thanks to @gedankenstuecke for pointing out I could log in and do that! And thank you to @jhdulaney for the review, I'll merge now.

Issue #827, opened 2018-08-10
Make project membership optionally public


madprime:
When a member has joined a project, that status is currently public on their profile. The only way to avoid this is to not be part of the project.

This is problematic in a variety of ways, and we think moving towards GDPR compliance implies giving members control over this information (about project membership status).

## Description
* Add a field to the project member model to track membership visibility
* Modify the project authorization template/view to ask/record membership visibility
* Add a view that allows a member to manage project membership visibility
* Honor visibility in a member profile page
* Honor visibility in the public data API endpoints (members by source, sources by member)

PR #833, opened 2018-08-16
Hide public membership in projects (#827)

jhdulaney:
## General Checkups

## Description
At this point, if the visible column in the private_sharing_datarequestprojectmember gets set to 'f', then the member's belonging to that particular project is hidden from the public.

This is a work in progress.

Still to do:

Front end for setting visibility
Tests

## Related Issue
#827


private_sharing/models.py
```diff
@@ -348,6 +390,7 @@ class DataRequestProjectMember(models.Model):
     joined = models.BooleanField(default=False)
     authorized = models.BooleanField(default=False)
     revoked = models.BooleanField(default=False)
+    visible = models.BooleanField(default=True)
```
madprime: Shouldn't this be on `DataRequestProjectMember`? (As it's specific to the member/project combination.)

jhdulaney: It is.

open_humans/templates/member/activity-management.html
```diff
@@ -148,6 +148,13 @@ <h3>Project has changed permissions!</h3>
     </div>
     {% endif %}
 
+    <div>
+        {% source_is_visible source as is_visible %}
+        Click here to toggle whether or not your membership in this project is
+        visible to outher members.  Hiding your membership leaves your data available:
```
madprime: minor typo :)

open_humans/templatetags/utilities.py
```diff
@@ -298,3 +299,37 @@ def add_string(a, b):
     Like `add` but coerces to strings instead of integers.
     """
     return str(a) + str(b)
+
+
+@register.tag
+def is_visible(parser, token):
```
madprime: I think this is acting as a rendering filter than returning a boolean, is that right? if so, I wonder about renaming it to `render_if_visible`?

private_sharing/models.py
```diff
@@ -74,6 +87,35 @@ def badge_upload_path(instance, filename):
...
 
+def project_membership_visible(user, project):
+    """
+    Determine if the user's membership in a project is visible or not.
+    """
+    project_id = id_label_to_project(project)
+    user_id = user_to_id(user)
+
+    if project_id is not None:
+        project = DataRequestProjectMember.objects.get(member_id=user_id,
```
madprime: I think this resulted in an error when I tried this out locally, I think due to an obsolete badge. Since badges can be out of sync, maybe this should have a `try` / `except DataRequestProjectMember.DoesNotExist`?

open_humans/templates/member/activity-management.html
jhdulaney: Fixed

private_sharing/models.py
```diff
@@ -74,6 +88,36 @@ def badge_upload_path(instance, filename):
...
+
+
+def toggle_membership_visibility(user, project, state):
+    """
+    Change the state of whether a member's data sharing is publicly visible or not.
+    """
+    project_id = int(project)
+    user_id = user_to_id(user)
+    state = bool(strtobool(state))
+    project = DataRequestProjectMember.objects.get(member_id=user_id,
```
madprime: Member IDs and user IDs are different. This function seems to be trying to accept both User and Member objects. Assuming what's actually generated with the user_to_id function is a User object ID, it isn't guaranteed to map to a Member ID (indeed, they're all off-by-one in our current database).

And... usually it's trivial to look one up from the other, `user.member` and `member.user`, so I think the function could just accept one (and then the variable name would be more accurate here) and in contexts where there's currently a member passed, pass the `.user` for that thing instead. (This also removes the complexity of the `user_to_id` code – which is confusing anyway, because "user" is being used to refer to things that aren't User objects, guarenteed to confuse I'm afraid! :confounded:

private_sharing/models.py
```diff
@@ -74,6 +75,34 @@ def badge_upload_path(instance, filename):
...
 
 
+def project_membership_visible(member, project):
+    """
+    Determine if the user's membership in a project is visible or not.
+    """
+    project_id = int(id_label_to_project(project).id)
+
+    if project_id is not None:
+        if DataRequestProjectMember.objects.filter(member_id=member.id).exists():
```
madprime: I believe you can just write `member=member` and `project=project`?

open_humans/api_views.py
```diff
@@ -98,15 +79,16 @@ class PublicDataUsersBySourceAPIView(APIView):
...
+                 .values('username', 'member__badges', 'id'))
         sources = defaultdict(list)
 
         for user in users:
             for badge in user['member__badges'] or []:
                 if 'label' not in badge:
                     continue
 
-                sources[badge['label']].append(user['username'])
+                if project_membership_visible(user, badge['label']):
```
madprime: Should this be `user.member`?

public_data/serializers.py
```diff
@@ -1,33 +1,50 @@
...
+    def to_representation(self, data):
+        ret = OrderedDict()
+        fields = self.get_fields()
+        query_params = dict(self.context.get('request').query_params)
+        source = getattr(data, 'source')
+        user_t = getattr(data, 'user')
+        usernames = []
+        if 'username' in query_params:
+            usernames = query_params['username']
+        visible = project_membership_visible(user_t, source)
```
madprime: should this be `user_t.member`?

open_humans/serializers.py
```diff
@@ -41,7 +42,6 @@ def get_sources(obj):
         if not hasattr(obj, 'member'):
             return []
 
-        sources = dict(get_source_labels_and_names_including_dynamic())
-
-        return sorted(badge['label'] for badge in obj.member.badges
-                      if 'label' in badge and badge['label'] in sources)
+        sources = (k for k, i in get_source_labels_and_names_including_dynamic()
+                   if project_membership_visible(obj, k))
```
madprime: should this be `obj.member`?

open_humans/templatetags/utilities.py
```diff
@@ -298,3 +324,37 @@ def add_string(a, b):
...
     return str(a) + str(b)
+
+
+@register.tag
+def render_if_visible(parser, token):
+    """
+    Is a member publicly sharing data but wishes that membership to not be public?
+    """
+    try:
+        tag, user_t, project_t = token.split_contents()
```
madprime: I think `user_t` is actually receiving a `member_t` and `project_t` might be more clearly named as a `source` or `source_label_t`? Naming should be consistent with the object being passed.

open_humans/templatetags/utilities.py
```diff
@@ -298,3 +324,37 @@ def add_string(a, b):
...
+class VisibleNode(template.Node):
+    def __init__(self, nodelist, user, project):
+        self.nodelist = nodelist
+        self.user = user
+        self.project = project
+
+    def render(self, context):
+        user = self.user.resolve(context)
+        project = self.project.resolve(context)
+        if project_membership_visible(user, project):
```
madprime: I think what's actually passed in the code at this point is a `member` despite it being called a `user` here – in which case the code variables should be renamed accordingly. I think it'd also be more clear to call it a `source` or `source_label` to be more clear that it's not a `DataRequestProject` object.

PR #835, opened 2018-08-16
Re-enable some of the API tests

madprime:
## General Checkups

## Description
Reenables some of the API tests.

To re-enable these, there's...
1. some updates to test data (datafiles and project) for testing purposes
2. adding an EmailAddress entry to test data, from django-user-accounts, for one of the test users
3. an update to our custom API authentication which I think may be accounting for some new behavior in django-rest-framework and/or django-oauth-toolkit that was introduced in the Django 2 upgrades

Also added a `unittest.skipIf` to skip tests that will break if an AWS S3 bucket isn't set up. The files in test data aren't actually present in the bucket/credentials I tested with, but the tests seem to be passing for me.

### DIFF
--- a/open_humans/fixtures/test-data.json
+++ b/open_humans/fixtures/test-data.json
@@ -177,6 +177,18 @@
     }
 },
 {
+    "model": "account.emailaddress",
+    "pk": 1,
+    "fields": {
+      "user": [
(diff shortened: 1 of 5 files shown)


madprime:
Thanks!

Issue #826, opened 2018-08-10
Webhook/notification upon project member withdrawal/deauthorization


madprime:
As part of updates towards GDPR compliance, we need a feature that notifies projects when a member has withdrawn/deauthorized, to enable them to act on requests to remove local copies of data.

## Description

1. View for all projects to see/download a list of project member IDs that have withdrawn/dauthorized & requested removal of data
2. This probably implies another field on a model, to keep a track of whether data removal was requested
3. Modification to the withdrawal/deauthorization page: ask member if they want the project to delete data it's received, if any
4. Webhook for OAuth2 Data Request Projects (optional)
5. Transactional email sent to projects (optional)

The "optional" items mean that a project can choose not to use/receive this. Our goal is to become GDPR compliant ourselves, and make it possible for projects to be GDPR compliant, but not require this.

The rationale for 1 & 2 is pragmatic: many projects are good faith, low resource efforts that do things manually and in batches. We can facilitate this by making it easy to find a full list of removal requests.

madprime:
Completed with #837 merged :tada:

PR #837, opened 2018-08-22
Add view to notify downstream projects of data removal

jhdulaney:
## General Checkups

## Description
Provides a view showing data that has been deauthorized and deleted so that downstream projects may easily consume this and themselves remove the data.

## Related Issue
#826


open_humans/signals.py
```diff
@@ -225,3 +226,23 @@ def email_confirmed_cb(email_address, **kwargs):
     Send a user a welcome email once they've confirmed their email address.
     """
     send_welcome_email(email_address)
+
+
+def send_withdrawel_email(project, slug):
```
madprime: typo, should be "withdrawal" throughout :)

I'm expecting to see a `@receiver` that's like `projectmember_pre_save`, that (a) sends an email to the project (maybe this should be (optionally) off if a webhook exists), and (b) POSTs data in a JSON format to the webhook/callback URL provided by the project (if provided) c.f. https://sendgrid.com/blog/whats-webhook/

data_import/views.py
```diff
@@ -48,3 +52,12 @@ def get_redirect_url(self, *args, **kwargs):
         access_log.save()
 
         return self.data_file.file_url_as_attachment
+
+
+class RemovedDataView(ListAPIView):
```
madprime: Is this publicly accessible? This might not matter if this API isn't needed in the first place. I didn't think we would have another API view, but instead support a webhook/callback if provided.

data_import/models.py
```diff
@@ -167,3 +167,14 @@ class TestUserData(models.Model):
     user = fields.AutoOneToOneField(settings.AUTH_USER_MODEL,
                                     related_name='test_user_data',
                                     on_delete=models.CASCADE)
+
+class RemovedData(models.Model):
```
madprime: I didn't expect to track withdrawals or data removal requests in the database, but to send an email and/or a notification to a webhook URL to the relevant project. (I think this would imply the migration above is unnecessary...)

One thing we may need to do, though, is add a field to `DataRequestProjectMember` which, upon deauthorization, records whether downstream data removal is requested. But then, I'm not sure whether all projects can support downstream data removal! (In which case we shouldn't offer? Does this also imply a parameter for `DataRequestProject`?) This brings up some design questions we might discuss in a meeting!

open_humans/signals.py
jhdulaney: English is hard.

madprime:
I'm going to expand on the feature as described in #826 ...

The first three items were...

> 1. View for all projects to see/download a list of project member IDs that have withdrawn/dauthorized & requested removal of data
> 2. This probably implies another field on a model, to keep a track of whether data removal was requested
> 3. Modification to the withdrawal/deauthorization page: ask member if they want the project to delete data it's received, if any

For (3) I expected a modification to `ProjectLeaveView` in `private_sharing.views` (and the associated template) that asks a member - as they leave a project - whether they want that project to delete any information they've received from Open Humans about the member. I think the `project_member.leave_project` would be extended to receive this boolean...

This would probably update... (2) a new boolean field in `DataRequestProjectMember` in `private_sharing.models`, possibly `erasure_requested`.

And for (1), not an API endpoint, but a page on the site available to a project manager. That is, a new template and view, similar to `OnSiteDataRequestProjectDetailView` and other views that use the `CoordinatorOnlyMixin` in `private_sharing.views`

> 4. Webhook for OAuth2 Data Request Projects (optional)

As noted in the issue, "optional" refers to this being an optional field for the project. The `OAuth2DataRequestProject` model should have a new text field (maybe `deauth_webhook`?) added where a project manager can provide a URL which receives a notification that a project member has withdrawn and requested data deletion. This is optionally blank.

If the project has defined this URL, then Open Humans should be sending a notification upon withdrawal. I had thought a signal might be needed, but I think that's overly complicated, this could be accomplished within the `leave_project` method of `DataRequestProjectMember`. A `POST` should be sent to the URL with some JSON format information, probably like `{ 'project_member_id': '12345678', 'erasure_requested': True}`

> 5. Transactional email sent to projects (optional)

The `DataRequestProject` model should have a new boolean field (maybe `deauth_email_notification`?) added which, if set, means the coordinator receives an email notification upon withdrawal and describes whether erasure has been requested.

As with for (4), I think this could be implemented in `leave_project`, it doesn't need to be a signal.

> 6. Erasure allowed (boolean setting for project)

(Just joking - this wasn't featured in original issue! I added it just now.)

As I walk through this, I realize that not all projects will want to offer erasure. Research may be exempt from GDPR, or not interested in EU regs. It's important that (3) does not _offer_ erasure to members unless the project wants to offer it! So, yet another boolean field I guess (sorry) on `DataRequestProject` (maybe `erasure_supported`?) that the project can set, and (3) only offers the option on withdrawal if it's set to `True`.

private_sharing/models.py
```diff
@@ -403,18 +421,42 @@ def random_project_member_id():
...
+
+    def leave_project(self, remove_datafiles=False, done_by=None, erasure_requested=False):
         self.revoked = True
         self.joined = False
         self.authorized = False
+        if erasure_requested:
+            self.erasure_requested = timezone.now()
         self.save()
 
+        if self.erasure_requested == None:
```
madprime: would it maybe be simpler to use `bool()`?

jhdulaney: Doh

private_sharing/utilities.py
```diff
@@ -50,3 +53,22 @@ def source_to_url_slug(source):
...
             return project.slug
+
+
+def send_withdrawal_email(project, slug):
+    """
+    Email a project to notify them that a member has withdrawn.
+    """
+
+    params = {
+        "withdrawn_url": full_url("/erase-member-data/{0}".format(project.slug)),
```
madprime: Ah, the URL as constructed here isn't working. Maybe passing the project_member to this template would make more sense..

And maybe you could use something like `reverse_lazy('direct-sharing:erase-member-data', kwargs={'slug': project_member.project.slug})`?

jhdulaney: huh, it seems to work for me?  I click on it, and it comes up.

private_sharing/migrations/0011_auto_20180911_2127.py
```diff
@@ -0,0 +1,28 @@
...
+
+    dependencies = [
+        ('private_sharing', '0010_datarequestprojectmember_visible'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='datarequestproject',
+            name='deauth_email_notification',
+            field=models.BooleanField(default=True, help_text='Receive emails when a member deauthorizes your project', verbose_name='Deauthorize email notifications'),
```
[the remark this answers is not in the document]
jhdulaney: good point

private_sharing/templates/email/notify-withdrawal.html
```diff
@@ -0,0 +1,31 @@
+<p>
+  Hello,
+</p>
+
+<p>
+  This email is to inform you that a member (0) has withdrawn their participation
```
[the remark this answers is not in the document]
jhdulaney: Can do

private_sharing/models.py
```diff
@@ -403,18 +421,39 @@ def random_project_member_id():
...
+
         if self.project.type == 'oauth2':
             application = self.project.oauth2datarequestproject.application
             AccessToken.objects.filter(user=self.member.user,
                                        application=application).delete()
             RefreshToken.objects.filter(user=self.member.user,
                                         application=application).delete()
+            if self.project.oauth2datarequestproject.deauth_webhook != '':
+                self.deauth_webhook(json_p)
+        if (self.project.deauth_email_notification and erasure_requested):
```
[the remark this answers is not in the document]
jhdulaney: I would think so?  I feel like otherwise might just be noise.  I guess it depends on how often people actually withdraw?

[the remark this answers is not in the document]
madprime: Well... the webhook operates regardless of data erasure request.

Also... studies that don't offer data erasure may still want to know about withdrawals, it's information the IRB cares about and it's plausible that it still affects their protocol somehow.

So my feeling isn't strong, but in the direction of notifying for any deauth event...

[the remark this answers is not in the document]
jhdulaney: Hrm, I wonder if the request erasure should only be available if they first select data deletion on OH?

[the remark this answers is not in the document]
jhdulaney: It seems to me that requesting downstream erasure and not deleting on OpenHumans means the downstream project would just pick the data back up again without some attention to ignoring it.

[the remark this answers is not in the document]
jhdulaney: I am going with sending an email no matter what.  I am also adding a column to the withdrawn requests page, Erasure requested, with a simple True or False associated with each project_member_id.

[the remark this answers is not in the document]
madprime: No, I think it's plausible to lose trust in the project -- but retain trust in Open Humans to keep a copy of one's data (and want to analyze it, and have the ability to share it with others).

private_sharing/views.py
```diff
@@ -543,3 +546,36 @@ def form_valid(self, form):
...
         return super(RemoveProjectMembersView, self).form_valid(form)
+
+class DataRequestProjectWithdrawnView(PrivateMixin, CoordinatorOnlyView,
+                                      ListView):
+    """
+    A view for coordinators to list members that have requested data removal.
+    """
+    model = DataRequestProject
+    paginate_by = 100
+    template_name = 'private_sharing/project-data-erase-view.html'
```
madprime: Maybe rename?

private_sharing/templates/private_sharing/project-data-erase-view.html
```diff
@@ -0,0 +1,43 @@
+{% extends 'base.html' %}
+
+{% load utilities %}
+
+{% block head_title %}Members requesting data erasure{% endblock %}
```
madprime: Update to "withdrawn" since it's all withdrawn members?

private_sharing/templates/email/notify-withdrawal.html
```diff
@@ -0,0 +1,32 @@
...
+<p>
+  This email is to inform you that a member, {{ project.project_member.project_member_id }},
+  has withdrawn their participation from your project.
+</p>
+<p>
+  For your convenience, we maintain a list of project members that have requested
+  withdrawal.  It can be found at {{ withdrawn_url }}.
+</p>
+<p>
+  {% erasure_requested_txt erasure_requested %}
```
madprime: I think I'd prefer an `{% if erasure_requested %}` block that only shows up if it's true. I'm thinking about projects that don't offer erasure in the first place - it's weird to get a statement like "erasure not requested" when it's never possibly true? Feels like it's mainly important to include when it's true, nothing needs to be said otherwise.

jhdulaney: Okay, I'll knock that out right quick.

PR #839, opened 2018-08-24
Add note about messaging prior to removal

danamlewis:
## Description
Making note about making sure to message members (if desired) before doing the removal, as project admin won't be able to message afterward.

Comment -  I am working on a template message for this to use. Would you like me to add a link to it here, or include it here? Can either include it on this page, or maybe link to it and store it on the wiki? (The latter might be my preference so it's easy for others to improve it directly in the wiki more quickly?)


danamlewis:
Template message added to the wiki, if you decide it's useful to link from on this page: https://github.com/OpenHumans/open-humans/wiki/Example-message-to-notify-members-prior-to-removal

PR #840, opened 2018-08-27
Add GDPR notes

madprime:
A page with notes on our GDPR compliance, a link to this in our signup form, and bumping up the age limit in our terms.

Because we're not aware of a _any_ individual using the site that's less than 18, let alone 16, I don't consider raising the age limit in the terms to be a major change.


open_humans/templates/base.html
```diff
@@ -46,6 +46,7 @@
     </noscript>
 
     {# For internal CSS stylesheets. #}
+    <!-- EXTRA CSS -->
```
jhdulaney: Extra CSS?

open_humans/templates/account/signup-form.html
```diff
@@ -11,21 +11,22 @@
...
     <p>
-      Your imported data will be private by default.
+      Imported data will be private by default. An account should be yours, or
+      someone for whom you are a legal guardian.
     </p>
+
     <p>
-      An account should represent an individual &ndash; your personal data, or
-      someone for whom you are a legal guardian.
+      This site is intended only for United States residents, but we
```
jhdulaney: Do intend to remain US-only indefinitely?  Otherwise, I do like the wording here.

open_humans/templates/pages/terms.html
```diff
@@ -152,9 +152,9 @@ <h3>Copyright</h3>
     </p>
     <h3>Kids</h3>
     <p>
-      Sorry, but Open Humans is not directed to people 12 and younger. If we
+      Sorry, but Open Humans is not directed to people 15 and younger. If we
       become aware that we have directly collected personal information from a
-      child under 13, we will delete that information.
+      child under 16, we will delete that information.
```
jhdulaney: maybe "... child under 16 without a parent or guardian's authorization"?

madprime: Authorization within US COPPA (which this partly covers) is a big headache, if I recall correctly. Which is probably why most websites just say "nope". Child data _can_ exist in Open Humans (and does, I expect), but the account has to be managed by a parent/guardian.

jhdulaney: Makes sense.

PR #845, opened 2018-08-30
Replace __unicode__ methods with __str__ and modernize some string formatting

### LINKED ISSUE #844: Fix model string representations
## General Checkups

## Description
Models have stopped displaying string representations as expected; see email to support. I think it might have to do with this -- seems to indicate that the `__unicode__` methods are obsolete/nonfunctional in Python 3? https://docs.djangoproject.com/en/1.11/topics/python3/#str-and-unicode-methods

## Example
The email to support mentioned above contained: `your account at <open_humans.settings.FakeSite object at 0x7f386bda6240>.` when it should contain the `name`.

jhdulaney:
## General Checkups

## Description
During the Python3 migration, I neglected to replace several __unicode__ methods with __str__ methods.  This has been done.  In addition, I modernized a few of the instances of string formatting.

## Related Issue
Fixes #844 

### DIFF
--- a/open_humans/models.py
+++ b/open_humans/models.py
@@ -26,13 +26,13 @@ def get_member_profile_image_upload_path(instance, filename):
     """
     Construct the upload path for a given member and filename.
     """
-    return 'member/%s/profile-images/%s' % (instance.user.id, filename)
+    return str('member/{0}/profile-images/{1}').format(instance.user.id, filename)
 
 def get_grant_project_image_upload_path(instance, filename):
     """
     Construct the upload path for an image for a ProjectGrant object.
     """
-    return 'grant-projects/%s/%s' % (instance.name, filename)
+    return str('grant-projects/{0}/{1}').format(instance.name, filename)
 
 def random_member_id():
     """
@@ -40,7 +40,7 @@ def random_member_id():
     any Member.
     """
     def random_id():
-        return '%08d' % random.randint(0, 99999999)
+        return str('{0:d}').format(random.randint(0, 99999999))
 
     member_id = random_id()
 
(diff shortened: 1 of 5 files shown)


open_humans/models.py
```diff
@@ -26,21 +26,21 @@ def get_member_profile_image_upload_path(instance, filename):
...
+    return str('grant-projects/{0}/{1}').format(instance.name, filename)
 
 def random_member_id():
     """
     Return a zero-padded string from 00000000 to 99999999 that's not in use by
     any Member.
     """
     def random_id():
-        return '%08d' % random.randint(0, 99999999)
+        return str('{0:d}').format(random.randint(0, 99999999))
```
gedankenstuecke: Are you sure this produces the right kind of padding? 

When running in an ipython shell I ended up with: 

```
In [52]: str('{0:d}').format(random.randint(0,99999999))
Out[52]: '91987'
```

open_humans/models.py
```diff
@@ -26,21 +26,21 @@ def get_member_profile_image_upload_path(instance, filename):
     """
     Construct the upload path for a given member and filename.
     """
-    return 'member/%s/profile-images/%s' % (instance.user.id, filename)
+    return str('member/{0}/profile-images/{1}').format(instance.user.id, filename)
```
gedankenstuecke: I noticed that we're giving the `{0}/{1}.format(foo,bar)` formatting in some places and in others only the `{}/{}` notation. Is there a reason why?

open_humans/models.py
jhdulaney: Ah, oops

open_humans/models.py
jhdulaney: No real reason.  I'll make them consistent.

gedankenstuecke: nice, thanks!

public_data/models.py
```diff
@@ -77,8 +77,8 @@ def __str__(self):
         if self.is_public:
             status = 'Public'
 
-        return str('{}:{}:{}').format(self.participant.member.user.username,
-                             self.data_source, status)
+        return str('{0}:{1}:{1}').format(self.participant.member.user.username,
```
gedankenstuecke: this should be `{0}:{1}:{2}` i think 😉

PR #846, opened 2018-08-31
Badgectomy

jhdulaney:
## General Checkups

## Description
This change does several things:
It removes dependency on badges for anything besides displaying images
Badges are now generated on the fly from database queries, not stored in a field in the Member model.  In turn, this means that badges no longer have to be manually generated.

## Tests
Currently, this passes local testing.  However, testing in staging may be warranted due to the large reach of this change.


open_humans/templatetags/utilities.py
```diff
@@ -357,3 +304,54 @@ def render(self, context):
...
+    return mark_safe(html)
+
+
+@register.simple_tag()
+def make_badge(project, badge_class='oh-badge'):
+    """
+    Return HTML for a badge.
+    """
+    if not isinstance(project, str):
+        print(project.badge_image)
```
madprime: is this leftover from debugging?

PR #852, opened 2018-09-10
Filter badges on approved

jhdulaney:
Quick fix to not show membership in unapproved projects

### DIFF
--- a/private_sharing/models.py
+++ b/private_sharing/models.py
@@ -81,12 +81,11 @@ def get_visible_user_projects(member):
     Returns a list of projects where membership is publicly
     visible.
     """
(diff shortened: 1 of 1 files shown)


private_sharing/models.py
```diff
@@ -81,12 +81,11 @@ def get_visible_user_projects(member):
     Returns a list of projects where membership is publicly
     visible.
     """
-    projects = []
     project_ids = DataRequestProjectMember.objects.filter(visible=True,
```
madprime: Rather than the list comprehension and additional database query below, could this database query filter on `project__approved=True`?

PR #860, opened 2018-09-13
Fix filtering members list on project

jhdulaney:
## General Checkups

## Description
Filtering members list by project was still trying to use the old (pre-badgectomy) code.  This fixes that.

## Related Issue
857

### DIFF
--- a/open_humans/member_views.py
+++ b/open_humans/member_views.py
@@ -6,8 +6,9 @@ import arrow
 
 from django.apps import apps
 from django.contrib import messages as django_messages
-from django.urls import reverse, reverse_lazy
+from django.db.models import Q
 from django.http import Http404, HttpResponseRedirect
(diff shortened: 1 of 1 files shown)


open_humans/member_views.py
```diff
@@ -87,8 +88,8 @@ def get_queryset(self):
             if not badge_exists:
                 raise Http404()
 
-            queryset = queryset.filter(
-                badges__contains=[{'label': filter_name}])
+            project = id_label_to_project(filter_name)
+            queryset = queryset.filter(datarequestproject=project)
```
madprime: oops! It looks to me like this ends up returning the `Member` that is `project.coordinator`?

Maybe you want something like `datarequestprojectmember__project=project` but ... I think also need to check that they're authorized/joined.

jhdulaney: o, oops

Issue #853, opened 2018-09-11
Restore "sort by badge count" to members list


madprime:
The update for badges (#846) seems to have introduced an inefficiency that's causing all member list pages to time out. This may be causing site-wide timeouts as servers get clogged on these requests.

1. Remove URL path for the list for now (it won't work anyway)
2. Fix the code so this works

jhdulaney:
Noting that the reason for this is that we are sorting the members list within Python, which causes the queryset to resolve.  The true solution will require doing the sort within the database.  This should be possible by annotating each member with a badge count that can be done entirely within the queryset.

PR #862, opened 2018-09-13
Restore sorting members list by number of badges

jhdulaney:
## General Checkups

## Description
Previously, sorting the members list was done within Python.  This is inefficient.  This PR now does the sorting within the database.

## Related Issue
#853


open_humans/member_views.py
```diff
@@ -97,12 +97,10 @@ def get_queryset(self):
...
 
-        projects = self.get_projects()
-        sorted_members = queryset.order_by('-id')
-        # Disabling for now until we can do this better
-#        sorted_members = sorted(queryset,
-#                                key=lambda m: projects.filter(member_id=m.id).count(),
-#                                reverse=True)
+        sorted_members = queryset.annotate(num_badges=Count('datarequestprojectmember__project',
+                                                            filter=Q(
+                                                                datarequestprojectmember__revoked=False
```
madprime: Will this list people "higher" -- but display less badges for them if they have a lot of "hidden" project memberships? I'm wondering if it's better to just count visible memberships.

jhdulaney: good point

jhdulaney:
I'm getting:

This branch has no conflicts with the base branch
Merging can be performed automatically.

Now, Codeclimate is complaining that we're at 26 lines ...

Issue #868, opened 2018-09-14
Members list displays badges for deauthed/withdrawn projects


madprime:
## Description
In the members list, currently members have badges shown for projects they have withdrawn/deauthorized.

The culprit is probably this line, which doesn't filter on authorized?
https://github.com/OpenHumans/open-humans/blob/master/open_humans/templatetags/utilities.py#L314

## Example
The listing for me below is showing me as a member of Nightscout and OpenAPS projects, which I joined at some point for testing purposed, but withdrew/deauthorized.

-----

![image](https://user-images.githubusercontent.com/82631/45571668-d3b0b080-b81b-11e8-990b-04b1a923d1cd.png)

jhdulaney:
Fixed #869

Issue #867, opened 2018-09-14
Public API "members by source" behaves weirdly on filter


madprime:
## Description
A filter applied to the "members by source" public API endpoint is behaving weirdly.

You can find examples of this behavior by looking at activity. There's a link related to the number of members that have connected the project, "Connected by X members".

## Example
For example, [23andMe Upload](https://www.openhumans.org/activity/23andme-upload/) currently says "Connected by 1024 members."

Clicking this link goes here: https://www.openhumans.org/api/public-data/members-by-source/?source=direct-sharing-128

Which produces something like this:

```
{
    "count": 27,
    "next": null,
    "previous": null,
    "results": [
        {},
        {},
        {},
        {
            "source": "direct-sharing-128",
            "name": "23andMe Upload",
            "usernames": [
                "member1",
                "member2",
                "member3",
                "member4",
                ...
                "member1022",
                "member1023",
                "member1024"
        },
        {},
        {},
        {}
    ]
}
```

I'm not sure what it used to do before recent updates, but it should probably not return a list of `{}` for all the projects that were filtered out?

jhdulaney:
Fixed #869

Issue #866, opened 2018-09-14
Can't hide membership for projects that aren't data sources


madprime:
It seems that I'm unable to hide project membership for any activities that aren't acting as data sources, the button isn't there.

e.g. screenshot below from [Genevieve Genome Report](https://www.openhumans.org/activity/genevieve-genome-report/)

---

![image](https://user-images.githubusercontent.com/82631/45569104-639e2c80-b813-11e8-9a20-6869939ffc8f.png)

jhdulaney:
Fixed #869

PR #869, opened 2018-09-17
Cute crawlers -- Fixes for small bugs

jhdulaney:
## Description
This has fixes for #866 #867 and #868

### DIFF
--- a/open_humans/serializers.py
+++ b/open_humans/serializers.py
@@ -1,6 +1,7 @@
 from collections import OrderedDict
 
 from django.contrib.auth import get_user_model
+from django.db import models
 from django.urls import reverse
 
 from rest_framework import serializers
@@ -52,6 +53,28 @@ class MemberDataSourcesSerializer(serializers.ModelSerializer):
         return sorted(sources)
 
 
+class NoNullSerializer(serializers.ListSerializer):
+    """
+    Override ListSerialzer's to_representation.
+    """
+
+    def to_representation(self, data):
+        """
+        List of object instances -> List of dicts of primitive datatypes.
+        Overrides ListSerializer to add for checking for empty items.
+        """
+        # Dealing with nested relationships, data can be a Manager,
+        # so, first get a queryset from the Manager if needed
+        iterable = data.all() if isinstance(data, models.Manager) else data
+
+        ret = []
+        for item in iterable:
+            repres = self.child.to_representation(item)
+            if repres != OrderedDict():
+                ret = ret + [repres]
+        return ret
+
+
 class DataUsersBySourceSerializer(serializers.ModelSerializer):
     """
     Serialize the members of a data source.
@@ -60,6 +83,7 @@ class DataUsersBySourceSerializer(serializers.ModelSerializer):
     class Meta:  # noqa: D101
         model = DataRequestProjectMember
         fields = ('id', 'project', 'visible')
+        list_serializer_class = NoNullSerializer
 
     def to_representation(self, data):
         ret = OrderedDict()
--- a/open_humans/serializers.py
+++ b/open_humans/serializers.py
@@ -55,7 +55,7 @@ class MemberDataSourcesSerializer(serializers.ModelSerializer):
 
 class NoNullSerializer(serializers.ListSerializer):
     """
-    Override ListSerialzer's to_representation.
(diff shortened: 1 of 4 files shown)


open_humans/serializers.py
```diff
@@ -52,6 +53,28 @@ def get_sources(obj):
         return sorted(sources)
 
 
+class NoNullSerializer(serializers.ListSerializer):
+    """
+    Override ListSerialzer's to_representation.
```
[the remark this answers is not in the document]
jhdulaney: It's cereal!

open_humans/templates/partials/visible-public-sharing.html
```diff
@@ -1,7 +1,5 @@
 <div class="">
-  {% if data_files %}
-
-    {% if user.member.public_data_participant.enrolled %}
+    {% if not project_member.revoked %}
```
madprime: I think this should be switched ... probably to also check `project_member.authorized` as done for the `render_user_badges` templatetag below.

Because `revoked` defaults to false on membership objects I'd worry a bit about accidentally having membership objects interpreted as being joined/authorized when they just failed to set `revoked` to True.

Regarding this whole page & inconsistency: `member/activity-management.html` template is checking `activity.is_connected`, which is determined by `personalize_activities_dict` in `common/activities.py`, which generally decides based on the `project.is_joined` boolean.

The `joined` and `authorized` distinction is an odd one that reflects a two-stage process for on-site activities that present a consent/terms = "joined", followed by an authorization = "authorized". In OAuth2 projects these should be synonymous. Since `authorized` is the second step, I think it's the one to we should use.

**All the code in `common/activities.py` is probably legacy.** (I'm taking this opportunity to make this point. :slightly_smiling_face: ) We used that stuff to combine "app-style projects" (legacy) with "model-style projects". The app style projects are gone now, the only non-standard thing that remains is "public data sharing".

open_humans/templatetags/utilities.py
```diff
@@ -312,7 +312,8 @@ def render_user_badges(member, badge_class='mini-badge'):
     Returns the html to render all of a member's badges.
     """
     projects = DataRequestProjectMember.objects.select_related('project').filter(
-                   visible=True, project__approved=True).filter(member_id=member.id)
+                   visible=True, project__approved=True, member_id=member.id,
```
madprime: is `member_id=member.id` the same as `member=member`?

jhdulaney: Oops, wasn't paying attention.

open_humans/templates/partials/visible-public-sharing.html
jhdulaney: Hrm, at this point, it'll probably be easier to just make a templatetag that outputs the html rather than having twenty nested ifs.  Instead, I'll just do a nice complex db query like I've started doing elsewhere.  That'll be easier in the templatetag than in the template itself.

As for common.activities, I sort of gathered that, but that'll be an involved project in itself to rip all that out.  There's still a few things that depend on that code that could be better done as direct db queries.

madprime: with the templates - agreed complex logic is rough! my understanding is that moving it into views is usually ideal (rather than templatetags).

and... I still think this line should change, even if it's a minor change :)

open_humans/views.py
```diff
@@ -458,6 +458,11 @@ def get_context_data(self, **kwargs):
                     in ['share_username', 'send_messages', 'share_sources',
                         'all_sources']]))
 
+        # Decided whether or not to show the toggle public visible button
+        show_toggle_visible_button = ((not project_member.revoked) and
+                                      project_member.authorized and
+                                      project_member.member.public_data_participant.enrolled)
```
madprime: wait, this shouldn't matter, should it? All members have the ability to hide membership?

It's just an added complication for members that can perform public data sharing (on how files get shared publicly) but all members have the ability to hide membership status...

jhdulaney: oh, so it should just test on authorized, then?

madprime: checking "not revoked" is also fine, it should also be true & doesn't hurt to include it.

but it shouldn't be filtering out people that didn't activate the "public data sharing" feature...

Issue #865, opened 2018-09-14
Replace django-user-accounts with django-allauth


jhdulaney:
This is the second step in providing social login (#863)

Currently, we use django-user-accounts for authentication.  However, this does not support external authentication providors.  Switching to django-allauth will allow us to add this feature in the future.  In addition, django-allauth provides some more flexibility in terms of user fields.  This should mean that some of the minor announces around unused fields in the database should go away.

PR #873, opened 2018-09-21
WIP Replace django-user-accounts with django-allauth

jhdulaney:
## General Checkups

## Description
So far, this is an in-place replacement for django-user-accounts.

## Related Issue
#865

## Note
Probably should figure out what to do with the tables we are no longer going to use.  We may also want to associate the existing account_account table which contains timezone and language information with each user, and continue to populate it with new users.

### DIFF
--- a/open_humans/account_views.py
+++ b/open_humans/account_views.py
@@ -1,18 +1,31 @@
 from django.contrib import messages as django_messages
-from django.contrib.auth import logout
-from django.urls import resolve, reverse_lazy
-from django.views.generic.edit import DeleteView
+from django.contrib.auth import logout, get_user_model
+from django.shortcuts import redirect
+from django.urls import resolve, reverse, reverse_lazy
+from django.views.generic.base import TemplateView
+from django.views.generic.edit import DeleteView, FormView
 
-from allauth.account.app_settings import EMAIL_VERIFICATION
-from allauth.account.utils import (complete_signup, user_email, user_username)
-from allauth.account.views import (LoginView,
-                           EmailView,
-                           SignupView)
+from allauth.account.app_settings import EMAIL_VERIFICATION, LOGIN_ON_PASSWORD_RESET
+from allauth.account.forms import default_token_generator
+from allauth.account.utils import (complete_signup,
+                                   perform_login,
+                                   url_str_to_user_pk,
+                                   user_email,
+                                   user_username)
+from allauth.account.views import (AjaxCapableProcessFormViewMixin,
+                                   LoginView,
+                                   EmailView,
+                                   PasswordResetView,
+                                   SignupView,
+                                   _ajax_response)
 
 from common.mixins import PrivateMixin
 from private_sharing.models import OnSiteDataRequestProject
 
-from .forms import MemberChangeEmailForm, MemberLoginForm, MemberSignupForm
+from .forms import (MemberChangeEmailForm,
+                    MemberLoginForm,
+                    MemberSignupForm,
+                    PasswordResetForm)
 from .models import Member
 
 
@@ -141,3 +154,62 @@ class UserDeleteView(PrivateMixin, DeleteView):
 
     def get_object(self, queryset=None):
         return self.request.user
+
+
+class ResetPasswordView(PasswordResetView):
+    """
+    Ooops, we've done lost our password, Martha!
+    """
+    form_class=PasswordResetForm
+
+
+class PasswordResetFromKeyView(FormView):
+    """
+    Let's get a new password!
+    """
+    template_name = 'account/password_reset_token.html'
+    form_class = PasswordResetForm
+
+    def _get_user(self, uidb36):
+        User = get_user_model()
+        try:
+            pk = url_str_to_user_pk(uidb36)
+            return User.objects.get(pk=pk)
+        except (ValueError, User.DoesNotExist):
+            return None
+
+    def dispatch(self, request, uidb36, key, **kwargs):
+        self.request = request
+        self.key = key
+        user = self._get_user(uidb36)
+        self.reset_user = user
+        if user is None:
+            return redirect('account-password-reset-fail')
+        token = default_token_generator.check_token(user, key)
+        if not token:
+            return redirect('account-password-reset-fail')
+        return super().dispatch(request, uidb36, key, **kwargs)
+
+    def get_context_data(self, **kwargs):
+        ret = super().get_context_data(**kwargs)
+        ret['action_url'] = reverse(
+            'account_reset_password_from_key',
+            kwargs={'uidb36': self.kwargs['uidb36'],
+                    'key': self.kwargs['key']})
+        return ret
+
+    def change_password(self, password):
+        user = self.reset_user
+        user.set_password(password)
+        user.save()
+        return user
+
+    def form_valid(self, form):
+        if form.is_valid():
+            self.change_password(form.clean_password())
+
+            if LOGIN_ON_PASSWORD_RESET:
+                return perform_login(
+                    self.request, self.reset_user,
+                    email_verification=EMAIL_VERIFICATION)
+            return redirect('home')
--- a/open_humans/account_views.py
+++ b/open_humans/account_views.py
@@ -3,9 +3,11 @@ from django.contrib.auth import logout
 from django.urls import resolve, reverse_lazy
 from django.views.generic.edit import DeleteView
 
-from account.views import (LoginView as AccountLoginView,
-                           SettingsView as AccountSettingsView,
-                           SignupView as AccountSignupView)
+from allauth.account.app_settings import EMAIL_VERIFICATION
+from allauth.account.utils import (complete_signup, user_email, user_username)
+from allauth.account.views import (LoginView,
+                           EmailView,
+                           SignupView)
 
 from common.mixins import PrivateMixin
 from private_sharing.models import OnSiteDataRequestProject
@@ -14,7 +16,7 @@ from .forms import MemberChangeEmailForm, MemberLoginForm, MemberSignupForm
 from .models import Member
 
 
-class MemberLoginView(AccountLoginView):
+class MemberLoginView(LoginView):
     """
     A version of account's LoginView that requires the User to be a Member.
     """
@@ -57,7 +59,7 @@ class MemberLoginView(AccountLoginView):
         return [self.template_name]
 
 
-class MemberSignupView(AccountSignupView):
+class MemberSignupView(SignupView):
     """
     Creates a view for signing up for a Member account.
 
@@ -67,28 +69,26 @@ class MemberSignupView(AccountSignupView):
 
     form_class = MemberSignupForm
 
-    def create_account(self, form):
-        account = super(MemberSignupView, self).create_account(form)
+    def form_valid(self, form):
+        # By assigning the User to a property on the view, we allow subclasses
+        # of SignupView to access the newly created User instance
+        self.user = form.save(self.request)
 
-        # We only create Members from this view, which means that if a User has
-        # a Member then they've signed up to Open Humans and are a participant.
-        member = Member(user=account.user)
-
+        member = Member(user=self.user)
         member.newsletter = form.data.get('newsletter', 'off') == 'on'
         member.allow_user_messages = (
             form.data.get('allow_user_messages', 'off') == 'on')
-
         member.name = form.cleaned_data['name']
-
         member.save()
 
-        # this may not be necessary, but we do return the account object here
-        # so if it's consumed by anything that uses the member we'll want it to
-        # have the up to date version that includes the user's preferences and
-        # name
-        account.user.member = member
+        try:
+            return complete_signup(
+                self.request, self.user,
+                EMAIL_VERIFICATION,
+                self.get_success_url())
+        except ImmediateHttpResponse as e:
+            return e.response
 
-        return account
 
     def generate_username(self, form):
         """Override as Exception instead of NotImplementedError."""
@@ -97,7 +97,7 @@ class MemberSignupView(AccountSignupView):
         )
 
 
-class MemberChangeEmailView(PrivateMixin, AccountSettingsView):
+class MemberChangeEmailView(PrivateMixin, EmailView):
     """
     Creates a view for the current user to change their email.
 
--- a/open_humans/forms.py
+++ b/open_humans/forms.py
@@ -1,15 +1,20 @@
-from account.forms import (
+from allauth.account.adapter import get_adapter
+from allauth.account.app_settings import AuthenticationMethod, AUTHENTICATION_METHOD
+from allauth.account.forms import (
+    AddEmailForm,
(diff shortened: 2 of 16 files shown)


open_humans/templatetags/utilities.py
```diff
@@ -361,3 +364,15 @@ def make_badge(project, badge_class='oh-badge'):
 @register.simple_tag()
 def template_bool(item):
     return bool(item)
+
+
+@register.simple_tag(takes_context=True)
+def get_next_url(context):
+    next = context['request'].GET.get('next', None)
+    if next:
+        return next
```
madprime: There's something complicated going on here when the URL target we originally wanted had parameters. If this happens, the `next` parameter needs to be encoded -- otherwise, everything after the first `&` in the original URL target is lost.

This does actually happen, potentially quite often, during an OAuth2 authorization for a project.

The answer seems likely to involve using `urllib.parse.encode` at some point in this process? I'm not sure if it should be here (maybe?) and/or if some additional decode also needs to be done on login.

open_humans/models.py
```diff
@@ -140,6 +140,8 @@ class Member(models.Model):
         unique=True,
         default=random_member_id)
     seen_pgp_interstitial = models.BooleanField(default=False)
+    password_reset_redirect = models.CharField(max_length=30, default='',
```
madprime: this produced an error for me on my first test :slightly_smiling_face: URLs can be quite long. maybe this should even be a TextField?

open_humans/account_views.py
```diff
@@ -141,3 +174,149 @@ def delete(self, request, *args, **kwargs):
...
+        # (end NOTE)
+        ret.update({"login_url": login_url})
+        return ret
+
+    def post(self, request, *args, **kwargs):
+        request = super().post(request, *args, **kwargs)
+        return request
+
+
+class PasswordResetFromKeyView(FormView):
```
madprime: Same question here - is there some reason not to inherit from the equivalent in `allauth`?

open_humans/account_views.py
```diff
@@ -141,3 +174,149 @@ def delete(self, request, *args, **kwargs):
...
+            member.password_reset_redirect = '' # Clear redirect from db
+            member.save()
+            messages = {
+                'settings_updated': {
+                    'level': django_messages.SUCCESS,
+                    'text': 'Password successfully reset.'},}
+            return redirect(unquote_plus(next_url))
+
+
+class PasswordChangeView(PasswordChangeView):
```
madprime: (As mentioned elsewhere, should avoid repeating the parent class name.)

open_humans/account_views.py
```diff
@@ -141,3 +174,149 @@ def delete(self, request, *args, **kwargs):
...
+    """
+    template_name = 'account/password_change.html'
+    form_class = ChangePasswordForm
+    success_url = reverse_lazy('my-member-settings')
+    def form_valid(self, form):
+
+        return super().form_valid(form)
+
+
+class ConfirmEmailView(ConfirmEmailView):
```
madprime: (As mentioned elsewhere, should avoid repeating the parent class name.)

open_humans/account_views.py
```diff
@@ -141,3 +174,149 @@ def delete(self, request, *args, **kwargs):
...
+class PasswordChangeView(PasswordChangeView):
+    """
+    Change the password
+    """
+    template_name = 'account/password_change.html'
+    form_class = ChangePasswordForm
+    success_url = reverse_lazy('my-member-settings')
+    def form_valid(self, form):
+
+        return super().form_valid(form)
```
madprime: ???

open_humans/account_views.py
```diff
@@ -141,3 +174,149 @@ def delete(self, request, *args, **kwargs):
...
+
+
+class ConfirmEmailView(ConfirmEmailView):
+    """
+    Subclass ConfirmEmailView to set the user email.
+    """
+
+    def post(self, *args, **kwargs):
+        """
+        Replace allauth's ConfirmEmailView.post because we want to do a few more
```
madprime: If I'm reading correctly, I think the only thing this is needed for is the email deletions - does that seem right? (see below regarding login on confirmation)

open_humans/account_views.py
```diff
@@ -141,3 +174,149 @@ def delete(self, request, *args, **kwargs):
...
+        except AttributeError:
+            pass # If someone tries to use an expired or incorrect key,
+                 # let allauth's error handling handle it
+
+        django_messages.add_message(
+            self.request,
+            django_messages.SUCCESS,
+            'account/messages/email_confirmed.txt',
+            {'email': confirmation.email_address.email})
+        resp = self.login_on_confirm(confirmation)
```
madprime: it seems like this could instead be configured with `ACCOUNT_LOGIN_ON_EMAIL_CONFIRMATION`?

open_humans/forms.py
```diff
@@ -183,3 +256,26 @@ def send_mail(self, sender, receiver):
                   sender.member.primary_email.email,
                   [receiver.member.primary_email.email],
                   html_message=html)
+
+
+class ResetPasswordForm(ResetPasswordForm):
```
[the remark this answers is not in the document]
jhdulaney: Good point, did that thing

open_humans/account_views.py
jhdulaney: Fixed

open_humans/account_views.py
jhdulaney: Fixed

open_humans/account_views.py
```diff
@@ -141,3 +174,149 @@ def delete(self, request, *args, **kwargs):
 
     def get_object(self, queryset=None):
         return self.request.user
+
+
+class ResetPasswordView(FormView):
```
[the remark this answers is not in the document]
jhdulaney: I actually don't remember what I was doing there.  I deleted basically all of that and properly subclassed the allauth class.

open_humans/account_views.py
jhdulaney: In this case, allauth tries to do something overly clever that doesn't quite work.

open_humans/account_views.py
```diff
@@ -67,28 +99,26 @@ class MemberSignupView(AccountSignupView):
...
         member.name = form.cleaned_data['name']
-
         member.save()
 
-        # this may not be necessary, but we do return the account object here
-        # so if it's consumed by anything that uses the member we'll want it to
-        # have the up to date version that includes the user's preferences and
-        # name
-        account.user.member = member
+        try:
```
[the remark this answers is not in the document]
jhdulaney: Done

open_humans/account_views.py
jhdulaney: oops

open_humans/account_views.py
jhdulaney: Well, it also sets the new email as primary.

open_humans/account_views.py
jhdulaney: Deleted

open_humans/account_views.py
jhdulaney: Just super() now, and kept the db bits (delete and set primary)

open_humans/account_views.py
```diff
@@ -56,8 +81,15 @@ def get_template_names(self):
 
         return [self.template_name]
 
+    def post(self, request, *args, **kwargs):
+        ret = super().post(self, request, *args, **kwargs)
+        try:
+            return redirect(unquote_plus(ret.url))
```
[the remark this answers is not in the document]
jhdulaney: It get set at line 151 of allauth.account.utils

The problem is the way that I am encoding the return url sort of short circuits django's HttpResponseRedirect, so that is why I added the extra code.  I will comment to that affect.

open_humans/forms.py
```diff
@@ -32,17 +34,35 @@ def _clean_password(child_class, self_instance, password_field_name):
...
+    A subclass of django-allauth's form that checks user is a Member.
     """
 
     authentication_fail_message = ("Your password didn't match the " +
                                    'username or email you provided.')
 
     def clean(self):
         """Check that the user is a Member."""
-        cleaned_data = super(MemberLoginForm, self).clean()
+        cleaned_data = super().clean()
```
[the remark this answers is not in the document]
jhdulaney: Probably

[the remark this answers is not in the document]
jhdulaney: killed with fire

common/utils.py
```diff
@@ -126,3 +128,12 @@ def origin(string):
     Coerce an origin to 'open-humans' or 'external', defaulting to 'external'
     """
     return 'open-humans' if string == 'open-humans' else 'external'
+
+
+def get_redirect_url(request):
+    redirect_to = get_request_param(self.request, 'next')
+    if redirect_to:
+        url = is_safe_url(unquote_plus(redirect_to), allowed_hosts=None)
```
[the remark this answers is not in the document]
jhdulaney: Might could mommick running locally ...

I mean, I very easily could get my dev setup to do certs, but, that is an added burden on everyone else.

open_humans/forms.py
```diff
@@ -82,13 +113,36 @@ def clean_password_new(self):
         return _clean_password(ChangePasswordForm, self, 'password_new')
 
 
-class PasswordResetTokenForm(AccountPasswordResetTokenForm):
+class PasswordResetForm(forms.Form):
     """
     A subclass of account's PasswordResetTokenForm that checks password length.
```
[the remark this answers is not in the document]
jhdulaney: The mapping is just a bit different from our existing template.  I could change the template, but I don't think it really matters?

I will update the docstring.

common/utils.py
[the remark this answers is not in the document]
madprime: oh, that's a point. there may be a settings to check, but then it's getting complex. and OH will redirect to https anything that hits it anyway so I can't actually imagine what attack could actually occur. let's forget it. :)

open_humans/account_views.py
```diff
@@ -141,3 +162,112 @@ def delete(self, request, *args, **kwargs):
 
     def get_object(self, queryset=None):
         return self.request.user
+
+
+class ResetPasswordView(AllauthPasswordResetView):
+    """
+    Ooops, we've done lost our password, Martha!
```
madprime: Hah. But, before wrapping this up, I'd like to check that docstrings are being used in a way that helps us later -- to help us and others understand the code that's here in the future. The sorts of questions I raised during review, anticipating what another developer (or "future you", hah) might wonder about. A standard approach: why a subclass was created (or _wasn't_ used - if it was redone from scratch that's very noteworthy), what it's doing differently. :)

e.g. `"""Subclass allauth to use a custom form that preserves a "next" parameter"""`

open_humans/forms.py
```diff
@@ -69,11 +73,19 @@ class MemberSignupForm(AccountSignupForm):
     class Meta:  # noqa: D101
         fields = '__all__'
 
+    def clean(self):
+        super().clean()
+
+        if 'terms' not in self.cleaned_data:
+            self.add_error('terms', _('You must accept our terms of service.'))
```
madprime: (minor) we use "terms of use" instead of "terms of service"

but also... I'm not sure I see this phrase in the form when I try using it?

jhdulaney: Ah, oops, yeah, this is handled in js.

open_humans/settings.py
```diff
@@ -450,6 +422,7 @@ def __str__(self):
 AUTHENTICATION_BACKENDS = (
     'oauth2_provider.backends.OAuth2Backend',
     'django.contrib.auth.backends.ModelBackend',
```
madprime: maybe add above this:
`    # ModelBackend before allauth + our User -> iexact email/username login`

open_humans/settings.py
```diff
@@ -386,14 +350,22 @@ def __str__(self):
...
-ACCOUNT_LOGIN_REDIRECT_URL = LOGIN_REDIRECT_URL
-ACCOUNT_OPEN_SIGNUP = to_bool('ACCOUNT_OPEN_SIGNUP', 'true')
-ACCOUNT_PASSWORD_MIN_LEN = 8
-ACCOUNT_SIGNUP_REDIRECT_URL = 'home'
-ACCOUNT_HOOKSET = 'open_humans.hooksets.OpenHumansHookSet'
-ACCOUNT_EMAIL_CONFIRMATION_AUTHENTICATED_REDIRECT_URL = 'home'
-ACCOUNT_EMAIL_CONFIRMATION_ANONYMOUS_REDIRECT_URL = 'home'
-ACCOUNT_USE_AUTH_AUTHENTICATE = True
+ACCOUNT_AUTHENTICATED_LOGIN_REDIRECTS = True
+# Due to our user model, this basically gets ignored
```
madprime: maybe more explanation like ...?
`# currently ignored due to custom User and ModelBackend (see above)`

open_humans/account_views.py
```diff
@@ -114,11 +132,14 @@ class MemberChangeEmailView(PrivateMixin, AccountSettingsView):
...
-        return super(MemberChangeEmailView, self).get_success_url(
-            *args, **kwargs)
+    def post(self, request, *args, **kwargs):
+        form = self.get_form()
+
+        if form.is_valid():
+            ret = self.form_valid(form)
+        else:
+            return self.form_invalid(form)
+        return ret
```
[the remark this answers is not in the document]
jhdulaney: Done

open_humans/account_views.py
```diff
@@ -97,14 +89,14 @@ def generate_username(self, form):
...
     Creates a view for the current user to change their email.
 
-    This is an email-only subclass of account's SettingsView.
+    Uses our own email change template.
     """
 
-    form_class = MemberChangeEmailForm
+    form_class = AllauthAddEmailForm
     template_name = 'member/my-member-change-email.html'
     success_url = reverse_lazy('my-member-settings')
```
madprime: Despite this here, it seems that submitting this form returns the user to the same form, instead of going to `my-member-settings`.

jhdulaney: fixed

open_humans/account_views.py
```diff
@@ -114,11 +106,18 @@ class MemberChangeEmailView(PrivateMixin, AccountSettingsView):
...
+        form = self.get_form()
+        email = str(form.data['email'])
+        emailaddress = EmailAddress.objects.filter(email=email)
+        if emailaddress.count() == 1:
+            if emailaddress.first().primary == False:
+                emailaddress.delete()
+
+        if form.is_valid():
+            self.form_valid(form)
+            return super().post(request, *args, **kwargs)
```
madprime: Why not `return self.form_valid(form)`?

AIUI the output of that is expected to be a response, and I think the parent is ultimately calling on the following via inheritance from `FormView`: https://github.com/django/django/blob/master/django/views/generic/edit.py#L57

jhdulaney: good point.  I still think that I should call super, though, just to get things like get_success_url() called.

jhdulaney: oh, you're right.  Killing the super()

Issue #875, opened 2018-10-03
Public Data API Server Error


dnvrdave:
## General Checkups
- [x] Have you checked that there isn't already an existing issue that describes what you report below?
- [x] Have you checked that there isn't already an open pull requests for this issue/update/change?

## Description

I get an error on the Public Data API when clicking this link:
“There are also two ways to see which users have which data sources:”
https://www.openhumans.org/api/public-data/sources-by-member/

## Example

This message pops up and it logs me out:
Server error
Sorry, your request resulted in a server error. The error has been logged and we'll investigate it.

madprime:
Fixed by #876 - thanks @dnvrdave & @jhdulaney !

PR #876, opened 2018-10-03
Sources by member broken

jhdulaney:
## General Checkups

## Description
Previously was not doing a check for the existence of a particular datarequestprojectmember with sufficient restraints.  That is no longer the case.

## Related Issue
Fix for  #875

### DIFF
--- a/open_humans/settings.py
+++ b/open_humans/settings.py
@@ -367,6 +367,11 @@ ACCOUNT_USERNAME_BLACKLIST = ['admin',
                               'moderator',
                               'openhuman']
 
+# We want CREATE_ON_SAVE to be True (the default) unless we're using the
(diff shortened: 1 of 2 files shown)


open_humans/settings.py
```diff
@@ -207,6 +207,7 @@ def __str__(self):
     'allauth',
     'allauth.account',
     'allauth.socialaccount',
+    'allauth.socialaccount.providers.google',
```
madprime: This looks like an accidental inclusion from the social login work. For clarity, I think it shouldn't be included here. :)

jhdulaney: oh, woops

jhdulaney: Fixed

PR #878, opened 2018-10-08
New hack to add template dir for a bulk email

madprime:
## Description
the `bulk_email` management command used to specific a custom template directory on the fly for rendering the emails. this is no longer available. This update works around the issue by allowing someone to specify an environment variable for an additional template directory (which contains the templates used for the email that will be sent)

### DIFF
--- a/open_humans/management/commands/bulk_email.py
+++ b/open_humans/management/commands/bulk_email.py
@@ -73,7 +73,7 @@ class Command(BaseCommand):
         email_file = full_path(options['email_file'])
 
         email_data = []
-        with open(email_file, 'rb') as f:
+        with open(email_file, 'r') as f:
             csvreader = csv.reader(f)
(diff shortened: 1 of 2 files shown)


open_humans/management/commands/bulk_email.py
```diff
@@ -73,7 +73,7 @@ def handle(self, *args, **options):
         email_file = full_path(options['email_file'])
 
         email_data = []
-        with open(email_file, 'rb') as f:
+        with open(email_file, 'r') as f:
```
jhdulaney: Curious as to why you are making file reading text-only?  It shouldn't really matter in this case.

madprime: I was getting this error, which was resolved by it:

`_csv.Error: iterator should return strings, not bytes (did you open the file in text mode?)`

Issue #863, opened 2018-09-14
Social login.


jhdulaney:
One of the big items on our usability improvements list is making it easier to create accounts and login with them.

To that end, we'd like to add the ability to create accounts and login using external identity provider services.

Currently, there are a couple of possibilities.

python-social-auth

python-social-auth (after a series of name changes was originally known as djang-social-auth)

Pros:
Already semi-integrated into the site

Cons:
Ugly code
Not very flexible
Last commit from May

django-allauth

Pros:
All in one auth solution
Active development
flexible (for example, for account creation we can range from zero questions when using an external identity provider to however many we want)

Cons:
Have to remove existing auth framework and provide a way to migrate over to the new method

Roll our own:

Pros:
Can get precisely what we want

Cons:
Still would have to remove or modify existing code
huge project
Future maintainability may be difficult

madprime:
:+1: I think that's best too. I guess the next bit of work may be to figure out these?
1. how hard is it to switch from django-user-accounts to django-allauth?
2. is it necessary or worth doing?

jhdulaney:
I think that it would be the correct thing to do to full switch to django-allauth.  I am currently unsure of the difficulty, though I think that it would mostly lay in maintaining a smooth migration.

I believe that I would like to proceed as follows:

Remove the existing python-social-auth code.  I would even go as far as to say that I would like this to be merged into master prior to proceeding.

Replace django-user-accounts with django-allauth

Impliment social login (which, at this point, should be trivial)

Once again, I'd like to do these as discrete steps, fully merged into master (and probably in prod) before proceeding.  I feel like this would minimize unexpected Gotchas.

madprime:
All three steps sound good to me! Let's go ahead and make an issue for each of those - I'll let you do the honors, since you made the plan. :)

jhdulaney:
Okay, now that's done.

I guess that we should decide on which login providers we would like to support.

Would it be better to whitelist or blacklist?

Full listing provided by allauth is here:
https://django-allauth.readthedocs.io/en/latest/installation.html

I think that we should whitelist.  Get the major players (FB, Googles, Twitter, Github), plus I'd suggest Gitlab, openid, persona, dropbox, baidu, bitbucket (both), and cern.

jhdulaney:
Oh, and, maybe reddit?

PR #880, opened 2018-10-08
Socialauth

jhdulaney:
## General Checkups

## Description
Activate django-allauth's social login feature

## Related Issue
#863


jhdulaney:
TODO still:
Better handling of existing email.  Will probably include a login link that includes the page where you add social accounts as the ?next= parameter
Tests

madprime:
@jhdulaney sorry, going ahead with review requires having app information for all three providers, and registering them as SocialApp objects in the admin interface. I can do that, but...

1. It would be very helpful to have some client secrets/IDs to use to register these? (kbfs works for me) or, if not that, some pointer on where to go to register for all three?

2. Related to that, did you set up any sort of generic "Open Humans" developer account with any/all of these providers? (I want to check if there's some team-level account information we should store in 1password.)

jhdulaney:
> @jhdulaney sorry, going ahead with review requires having app information for all three providers, and registering them as SocialApp objects in the admin interface. I can do that, but...
> 
>     1. It would be very helpful to have some client secrets/IDs to use to register these? (kbfs works for me) or, if not that, some pointer on where to go to register for all three?
> 
>     2. Related to that, did you set up any sort of generic "Open Humans" developer account with any/all of these providers? (I want to check if there's some team-level account information we should store in 1password.)

I can setup some Open Humans dev accounts.

madprime:
Some questions...

1. Facebook is asking me to re-enter my password, despite already being logged in to FB. I didn't expect this. Can this be avoided?

2. When FB returned me to the signup and I had an existing email -- why offer signup if we can already see the email is taken?

3. The Twitter connection doesn't ask for email address. Is that not an available option?

4. The Twitter connection also asks for permissions I don't think it needs - it's able to post tweets and update my profile. Is asking for these unavoidable?

And notes for myself -- some pages I might look at to make my own style/html code suggestions (i.e. don't worry about addressing these):
- /account/login/
- login modal
- /account/login/?socialsignup=true&next=/account/social/connections/

jhdulaney:
> Some questions...
> 
>     1. Facebook is asking me to re-enter my password, despite already being logged in to FB. I didn't expect this. Can this be avoided?

I have not encountered this, so I am unsure as to the conditions that trigger it.  I will investigate.

>     2. When FB returned me to the signup and I had an existing email -- why offer signup if we can already see the email is taken?

Yeah, that's probably a better way to do things.

>     3. The Twitter connection doesn't ask for email address. Is that not an available option?

I am somewhat unsure about this.  I have read that Twitter does not provide email addresses, but I have also read that they do (both in Twitter's own docs).  I'm going to have to investigate this a bit.

>     4. The Twitter connection also asks for permissions I don't think it needs - it's able to post tweets and update my profile. Is asking for these unavoidable?

The allauth docs don't mention any way to control what Twitter asks for, though allauth does provide a mechanism for that with Google and Facebook.

> And notes for myself -- some pages I might look at to make my own style/html code suggestions (i.e. don't worry about addressing these):
> 
>     * /account/login/
> 
>     * login modal
> 
>     * /account/login/?socialsignup=true&next=/account/social/connections/

jhdulaney:
For the bit where right now we are taken to the signup form instead of a login form, allauth has not yet implemented this.  I will work with upstream to get this done.

madprime:
Let's add comments to the HTML or other code that link to where assets come from

madprime:
Re @gedankenstuecke's note "when a user is trying to login via their username but only having used social login before it will throw a generic error message as there is no valid password" → I'm in favor of ignoring this edge case and letting the user reset the password. This should work, and my recollection is that this is how many/most sites end up handling the scenario.

gedankenstuecke:
Mh... The error message reads `Error: The username and/or password you specified are not correct.`. Do we feel that this prompts people to request a password reset? I feel the message is so vague that I might just doubt I put in the right username?

static/css/_pages.css
```diff
@@ -444,3 +444,57 @@ a.oh-badge {
     margin-top: 25px;
   }
 }
+
+.top-buffer {
```
madprime: The styles here have ignored that the way styles have been implemented in `_pages.css` was with page-specific scoping. Everything is nested within an id that starts with `#page-`, and this approach is being used to avoid styles that were only intended for a specific page (and might produce unintended consequences if the same class names are used elsewhere).

But because modals can be on any page, this approach probably should _not_ be used. New needed styles should be somewhere else, maybe `_buttons.css`.

open_humans/templates/socialaccount/signup.html
```diff
@@ -0,0 +1,154 @@
...
+{% load account socialaccount %}
+{% load utilities %}
+
+
+{% block head_title %}Create an Open Humans account{% endblock %}
+
+{% block panel_content %}
+
+<div class="well">
+  <p>Members must have a <strong>public username and public
```
madprime: The Balsamiq mock-ups have condensed language that hasn't been implemented here.

open_humans/templates/account/signup-form.html
```diff
@@ -98,10 +101,14 @@
 
     <div class="col-sm-8">
       <input type="password" class="form-control" name="password1"
-        id="signup-password" placeholder="Password"
+        id="password" placeholder="Password" data-toggle="password"
```
madprime: It's unclear to me why this id was renamed?

jhdulaney: Because I could easily use the library that provides the reveal password functionality.

madprime: You mean L109? Which could use `#signup-password` instead of `#password`, right?

jhdulaney: You're right, I don't recall why I did that.  I put it back.

open_humans/templates/account/login-modal.html
```diff
@@ -13,13 +16,31 @@ <h4 class="modal-title">Log in to Open Humans</h4>
...
+                  <button type="button" class="btn btn-default"
+                          data-dismiss="modal">Close</button>
+
+                  <input type="submit" form="login-form" id="login"
+                         class="btn btn-primary" value="Log in">
+                </div>
+              </div>
+            </form>
+          </div>
+          <div class="col-md-auto">
```
madprime: This appears to be a Bootstrap 4 class, but the site is using Bootstrap 3.

open_humans/templates/account/signup.html
```diff
@@ -1,30 +1,33 @@
...
-    {% include 'account/signup-form.html' %}
-  </div>
-</div>
 
-<div class="row">
-  <div class="col-xs-12 text-right">
-    <input type="submit" id="create-account" form="signup-form"
-      class="btn btn-primary" value="Create my account">
-  </div>
+<div class="col-md-auto">
```
madprime: This change removed a parent row div. Column divs should be inside row divs, per https://getbootstrap.com/docs/3.3/css/

open_humans/templates/account/signup-modal.html
```diff
@@ -11,24 +11,32 @@ <h4 class="modal-title">Create an Open Humans account</h4>
       </div>
 
       <div class="modal-body">
-        {% include 'account/signup-form.html' %}
+        <div class="col-md-auto">
```
madprime: In this case the div doesn't seem like this has any purpose. There's no row/column grid in the original version of this template, and this class is meaningless...

open_humans/templates/account/login-form.html
```diff
@@ -1,10 +1,8 @@
 {% load utilities %}
 
 {# Note - submit button not included. #}
-<form class="form-horizontal" role="form" method="POST"
-  action="{% url 'account_login' %}" id="login-form">
   {% csrf_token %}
-
+  <div class="row">
```
madprime: These additions are odd, I've not seen Bootstrap use rows within forms...

open_humans/urls.py
```diff
@@ -137,6 +141,11 @@
     path('account/login/oauth2/', views.OAuth2LoginView.as_view(),
          name='account-login-oauth2'),
 
+    # use our own template.
+    path('account/social/signup/',
+         account_views.SocialSignupView.as_view(),
+         name='SocialSignupView'),
```
madprime: There's inconsistency here in `name` declarations and what you're doing elsewhere.

The form for this page is POSTing to the URL name `socialaccount_signup`. This is defined by django-allauth to be this path... and that ends up working, as you've overridden the path django-allauth usually uses (but it will still happily point you to it).

But this `name` isn't being used at all, and is inconsistent with the naming scheme we've used...

open_humans/templates/socialaccount/signup.html
```diff
@@ -0,0 +1,154 @@
...
+
+  <p>
+    This site is intended only for United States residents, but we
+    strive to meet EU GDPR standards. <a href="{% url 'gdpr' %}">Read more
+      here.</a>
+  </p>
+</div>
+
+
+<form class="signup" id="signup_form" method="post"
```
madprime: @jhdulaney the class and id here seem to not be used anywhere else – is that correct?

jhdulaney: That template is a modification of the one that ships with allauth; that particular class and id comes from there.

open_humans/urls.py
```diff
@@ -137,6 +141,11 @@
...
+
     path('account/', include('allauth.urls')),
```
madprime: Examining this... it looks like it wasn't implemented correctly as a fallback here in the first place, these are wonky paths here.

image

jhdulaney:
Merged

PR #881, opened 2018-10-08
Add me

jhdulaney:
I guess I'm doing this thing


open_humans/templates/pages/about-staff.html
```diff
@@ -67,4 +67,20 @@ <h4 class="text-center">Mike Escalante<br>
     </div>
   </div>
 
+  <div class="row">
+    <div class="col-xs-4 visible-xs home-profile-div">
+      <img class="img-circle img-responsive img-center profile-pic-xs"
+        src="{% static 'images/mdulaney.jpg' %}" alt="Mairi Dulaney">
```
madprime: Looks good! but two requests...

1. the image needs to be square dimensions or close to it to display properly with these styles. (On inspection it seems bastian & I have not quite square images.)

2. please go ahead and put yourself above mike, we should list full time before part time :)

PR #883, opened 2018-10-15
Cleanup

jhdulaney:
Some code cleanup


jhdulaney:
I heard about this thing that finds dead code called Vulture and decided to play with it.

open_humans/forms.py
```diff
@@ -81,7 +80,7 @@ class ChangePasswordForm(AllauthChangePasswordForm):
     A subclass of account's ChangePasswordForm that checks password length.
     """
 
-    def clean_password_new(self):
+    def clean_password(self):
```
madprime: Sorry for the slow response on the PR. I started looking at the whole `_clean_password()` function here -- it was created to add some very basic "password strength" validation (i.e. a minimum length). This all seems to be obsolete.

1. It doesn't seem to be working? It's supposed to be imposing 8 characters minimum, and it's doing 6.
2. 6 is allauth's default in the absence of a password strength validator (a feature introduced in Django 1.9), see: https://github.com/pennersr/django-allauth/blob/master/allauth/account/app_settings.py#L147
3. We can set `ACCOUNT_PASSWORD_MIN_LENGTH = 8` and fix it to 8. This means adding three characters to this line (`LEN` -> `LENGTH`): https://github.com/OpenHumans/open-humans/blob/master/open_humans/settings.py#L364
4. And you can remove all this `_clean_password` override stuff that wasn't happening anyway - more code cleaned up :+1:

open_humans/models.py
```diff
@@ -88,10 +87,6 @@ def get_queryset(self):
                 .select_related('member')
                 .select_related('member__public_data_participant'))
 
-    def get_by_natural_key(self, username):
```
madprime: My understanding was that this in place to enable the case-insensitive username and email address lookup. Notably, upon login -- see how it's used in the Django `authenticate` backend here: https://github.com/django/django/blob/stable/2.1.x/django/contrib/auth/backends.py#L16

Removing this seems to be harmless now. The `ACCOUNT_AUTHENTICATION_METHOD` actually becomes relevant if this override is removed, and is resulting in the same behavior.

Sadly I think this might all indicate the custom `User` model is now unnecessary.

I don't think removing the custom User is in scope for this update (and it might be annoying and not worth it). But you should update all the docstrings to indicate what happened. :sweat: Don't just remove the claim that the custom models are enabling this look-up, explain that the model historically had this feature that's no longer present (i.e. the model might exist at this point for an obsolete reason).

jhdulaney:
Closing for now

Issue #886, opened 2018-10-19
Characters in usernames


madprime:
## General Checkups
- [x] Have you checked that there isn't already an existing issue that describes what you report below?
- [x] Have you checked that there isn't already an open pull requests for this issue/update/change?

## Description
I'm not sure how it was previously enforced, but I believe usernames used to only allow ASCII alphanumeric and underscore characters. Apparently this is no longer the case, as demonstrated by one new user that created an account with a username that matched their email.

This broke the main page, which was unable to run `reverse` for a `member-detail-direct` URL as the URL is expected to only contain the aforementioned characters. As a temporary fix, I set this user's username (approximately) to the `name` they selected during signup. (It's unclear whether this user intended their email to be public as a username, which is what they had done.)

A general fix should be implemented, either by rendering code like `urls.py` consistent with relaxed restrictions on characters in the username and/or re-imposing those restrictions on the accounts.

PR #888, opened 2018-10-19
make urls take non-ascii chars

jhdulaney:
## General Checkups

## Description
Using \w in the url patterns regex makes it accept non-ascii word chars.

## Related Issue
#886 

### DIFF
--- a/open_humans/urls.py
+++ b/open_humans/urls.py
@@ -155,7 +155,7 @@ urlpatterns = [
          member_views.MemberDataView.as_view(),
          name='my-member-data'),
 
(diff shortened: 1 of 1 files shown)


madprime:
As an added note (sorry to carp here) I don't think this would have worked for `@` characters and others anyway (so we'd still have the same problem), c.f. the Django validators including that character explicitly in the validator... https://github.com/django/django/blob/master/django/contrib/auth/validators.py#L10

madprime:
Closing in favor of #891 - it's possible to relax our username pattern but for now let's retain the existing behavior.

PR #892, opened 2018-10-19
fix error on login/signup

### LINKED ISSUE #890: Success at `/account/signup/` and `/account/login/` result in errors
## General Checkups

## Description
While the login or signup form in the modal works correctly from the main page, signup on the non-modal equivalents `/account/signup/` or `/account/login/` results in an error.

This seems to be related to a reverse attempted on the string `"None"`.

jhdulaney:
## General Checkups
- [ ] Have you checked that there aren't other open pull requests for the same issue/update/change?

## Description
Was not checking if the redirect urls were 'None' (yes, they are strings, not just None) or not.

## Related Issue
Fixes #890 

### DIFF
--- a/open_humans/account_views.py
+++ b/open_humans/account_views.py
@@ -4,6 +4,7 @@ from django.contrib import messages as django_messages
 from django.contrib.auth import logout, get_user_model
 from django.shortcuts import redirect
 from django.urls import resolve, reverse, reverse_lazy
+from django.urls.exceptions import NoReverseMatch
 from django.utils.http import is_safe_url
 from django.views.generic.edit import DeleteView, FormView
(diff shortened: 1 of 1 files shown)


open_humans/account_views.py
```diff
@@ -44,6 +44,8 @@ def post(self, request, *args, **kwargs):
         django's HttpResponseRedirect, which doesn't quite handle it correctly.
         """
         ret = super().post(self, request, *args, **kwargs)
+        if ret.url == 'None':
```
madprime: There's an `AttributeError` below. I had thought it was to catch a missing `url` attribute. Is that not the case?

PR #917, opened 2018-11-22
Do not allow authorizations with inactive oauth2 projects

jhdulaney:
## Description
While oauth2 projects would no longer show up anywhere on Open Humans, the external components of the project could still send people to be authorized.

## Related Issue
796

## Testing
  * tested authorizing with active oauth2 project
  * test authorizing with inactive oauth2 project
  * tested authorizing with empty client_id

### DIFF
--- a/private_sharing/views.py
+++ b/private_sharing/views.py
@@ -7,6 +7,8 @@ from django.urls import reverse, reverse_lazy
 from django.views.generic import (CreateView, DetailView, FormView, ListView,
                                   TemplateView, UpdateView, View)
 
+from oauth2_provider.models import Application
+
 from common.activities import personalize_activities_dict
 from common.mixins import LargePanelMixin, PrivateMixin
 from common.views import BaseOAuth2AuthorizationView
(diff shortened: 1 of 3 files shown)


private_sharing/views.py
```diff
@@ -268,9 +270,18 @@ class AuthorizeOAuth2ProjectView(ConnectedSourcesMixin, ProjectMemberMixin,
     template_name = 'private_sharing/authorize-oauth2.html'
 
     def dispatch(self, *args, **kwargs):
+        client_id = self.request.GET.get('client_id', None)
```
madprime: Is `self.get_object` not going to work here? Because this seems redundant to that...

madprime: sorry, I guess the thing to look at is the self.application property https://github.com/OpenHumans/open-humans/blob/master/common/views.py#L35 ...this code seems unnecessary (see below)...

private_sharing/views.py
```diff
@@ -268,11 +270,19 @@ class AuthorizeOAuth2ProjectView(ConnectedSourcesMixin, ProjectMemberMixin,
...
+            project = OAuth2DataRequestProject.objects.get(application=app)
+        else:
+            raise Http404
+
         if not self.application.oauth2datarequestproject:
             raise Http404
 
-        return super().dispatch(
-            *args, **kwargs)
+        if not project.active:
```
madprime: why not `self.application.oauth2datarequestproject.active`?

Issue #898, opened 2018-10-25
ActivityFeed should not display events where membership is hidden


madprime:
## General Checkups
- [x] Have you checked that there isn't already an existing issue that describes what you report below?
- [x] Have you checked that there isn't already an open pull requests for this issue/update/change?

## Description
The ActivityFeed appears to display a membership join event despite membership being hidden.

This should occur even if membership was hidden _after_ the join event is logged. (A user might not realize the membership was shared until after seeing it in ActivityFeed information, at which point they hide it. It's safe to assume they don't want the join event to be visible either.)

## Related Issue(s)
#897 

## Example
I discovered this issue as well when investigating #897. Hiding membership after joining didn't hide the related event in the activity feed.

PR #922, opened 2018-11-29
Hide events involving projects in Activity Feed if the member has selected to hide membership

jhdulaney:
## Description
With this PR, all events relating to projects no longer show up in the Activity Feed if the user has set membership in that public to hidden.

## Related Issue
#898 

## Testing

    * Confirmed that joining project still shows up as a New Member entry
    * If you join a project and choose to hide membership, joining does not show up in Activity Feed.
    * If join project but leave membership visible, the joining shows up
    * If you join, leave membership visible, and subsequently set membership to hidden, you no longer show up

### DIFF
--- a/open_humans/views.py
+++ b/open_humans/views.py
@@ -240,9 +240,10 @@ class HomeView(NeverCacheMixin, SourcesContextMixin, TemplateView):
         """
         Lists the 12 most recent actions by users.
         """
-        # Here we must use raw sql because the ORM is not quite able to do
-        # the complex 'where' we have here, or at least that's the consensus
-        # in #django on freenode
+        # Here we must use raw sql because the ORM is not quite able to take
+        # a queryset, look up two separate foreign keys in two separate models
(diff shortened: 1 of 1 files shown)


open_humans/views.py
```diff
@@ -234,7 +237,20 @@ def get_recent_blogposts():
 
     @staticmethod
     def get_recent_activity():
-        recent = ActivityFeed.objects.order_by('-timestamp')[0:12]
+        """
+        Lists the 12 most recent actions by users.
+        """
+        # Here we must use raw sql because the ORM is not quite able to do
+        # the complex 'where' we have here, or at least that's the consensus
+        # in #django on freenode
```
madprime: Django can't handle taking one object query, then looking up two separate fields from it in an unrelated model, jointly (in this case member_id and project_id).

madprime:
Sorry for slowness – and thank you for the updated explanation, it's much clearer to me!

PR #933, opened 2018-12-19
Don't require ssl for local

jhdulaney:
## Description
Due to a 'get the fix out the door' thing when heroku started requiring ssl for db connections, did not properly account for running locally (e.g. without ssl).  This commit fixes that oversight.

## Related Issue
Not officially reported

## Testing
ran the server with manage.py runserver
ran automated tests


jhdulaney:
resolved otherwise

Issue #918, opened 2018-11-27
Exchange member endpoint is slow


madprime:
## Description
The API endpoint for a project getting member-specific information is very slow. This is leading to Heroku alerts about long responses.

The URL for the endpoint is `/api/direct-sharing/project/exchange-member/`.

## Example
The following situation reproduces a long response...

* Running the site locally
* Using read-only access to the production database
* Using `requests` to call this endpoint with an access token corresponding to the in-development `PersonalDataAPI` project (project id: 187) and member `madprime`

gedankenstuecke:
This was uncovered through the `Overland` project which (accidentally) amassed 300+ small files for some members. 

With so many files the bottleneck probably is the S3 signing of download links returned when doing the `exchange-member`. In the most extreme case observed this leads to a page load time of >30 seconds, which in turn is terminated by Heroku.

gedankenstuecke:
Added the `affects-users` label as users will be unable to load the activity page in such cases as well due to the heroku timeout.

madprime:
In addition to the API endpoint, activity pages with premade S3 links are also failing when there's too many files. We think we should extend the public file URL method to work for private files, using some sort of token authorization. This URL then redirects to an S3 link generated on the fly, one file at a time.

jhdulaney:
Closed via #935 and #937

PR #935, opened 2018-12-21
Serialize datafiles with a url to our server and not aws

jhdulaney:
## Description
Previously, when /api/direct-sharing/project/members/ and /api/direct-sharing/project/exchange-member/ where hit, we would provide direct links to files on aws.  This would bog down quite a bit if a member had many files, as if you have n files you have n requests to aws to construct a url.  However, all of our publicly facing apis (ie, not requiring some form of auth to view) simply are serialized with a link to e.g. /data-management/datafile-download/424242/.  The AWS link would not be generated until this link that we provide was actually accessed.

This change makes it true across the board.

## Related Issue
#918 

## Testing
    * reproduced bug
    * bug no longer seems to be evidenced using local copy of the prod db
    * also tested /api/direct-sharing/project/members/ as that is affected by this change.


data_import/serializers.py
```diff
@@ -1,3 +1,5 @@
+from collections import OrderedDict
```
gedankenstuecke: Do we need this? I don't see it being used anywhere?

data_import/views.py
```diff
@@ -34,7 +34,17 @@ def get(self, request, *args, **kwargs):
         if unavailable:
             return HttpResponseForbidden('<h1>This file is unavailable.</h1>')
 
-        return super(DataFileDownloadView, self).get(request, *args, **kwargs)
+        query_key = request.GET.get('key', None)
+        if not self.data_file.is_public:
+            if not query_key:
+                return HttpResponseForbidden('<h1>No key provided.</h1>')
```
gedankenstuecke: Do we expect people to see these? Wondering whether it might be worthwhile to throw a proper error with the response instead of giving it a human-readable message? Or do I miss how HttpResponseForbidden works? :)

jhdulaney: Well, we could go with this:
https://docs.djangoproject.com/en/2.1/ref/views/#the-403-http-forbidden-view

I was adding to the already existing behaviour.  I feel it's nice to say why things are failing, which is the nice thing about returning something with HttpResponseForbidden.

Honestly, I could go either way.

jhdulaney: Related docs on HttpResponseForbiden:
https://docs.djangoproject.com/en/2.1/ref/request-response/#django.http.HttpResponseForbidden

If you think we'd be better off throwing a generic 403, then probably rewriting the entire method to be consistent would be a good idea.

PR #938, opened 2019-01-02
Edit logic to allow file download if user is logged in, regardless of key state.

jhdulaney:
## Description
This change:
* makes the DataFileDownloadView go from subclassing RedirectView to View
* Checks that if a user is authorized, and if so provide the download redirect to s3
* Replace calling super with a call to a function that logs file download and actually provides the redirect link

## Related Issue
N/A

## Testing
<!---
Please list of any testing you have done for this work.

This is not focused on automated tests (although those are nice).
Report anything, from basic "ran the code locally" to a design review.

For example, the following might be reported for a PR fixing a bug in
handling the "title" field a submitted form:
    * Can download via key from the member exchange endpoint
    * can download without key from the activity page
    * can download public files without key whether logged in or out
    * can download private files with key whether logged in or out

### DIFF
--- a/data_import/views.py
+++ b/data_import/views.py
@@ -1,8 +1,8 @@
 import logging
 
 from django.contrib.auth import get_user_model
-from django.http import HttpResponseForbidden
-from django.views.generic import RedirectView
+from django.http import HttpResponseForbidden, HttpResponseRedirect
+from django.views.generic import View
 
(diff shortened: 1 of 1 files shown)


jhdulaney:
I note that the RedirectView's get method was including a bunch of logic for config that we don't need.

PR #940, opened 2019-01-03
Better logging when files are accessed.

jhdulaney:
## Description
This change increases the information stored when a datafile key is created, such as the ip address causing the creation, project id, and access token.

This change also stores all this information, plus key creation timestamp and the key itself, in the datafile access log whenever a datafile is accessed.

## Related Issue
n/a

## Testing
* passed automated testing
* tested basic file download functionality
  - download of public files while logged out
  - download of public files while logged in
  - download of private files while logged in
  - download of private files while logged out but with key
* verified that each of the above tests produced the correct logs


data_import/models.py
```diff
@@ -117,18 +124,23 @@ def file_url_as_attachment(self):
         """
         return self.file.storage.url(self.file.name)
 
-    @property
```
madprime: It's funny, I wonder if this was intended to be used in templates (the way `download_url` is! But I don't think it ever was. Removing this seems to be fine.

data_import/serializers.py
```diff
@@ -8,10 +11,33 @@ class DataFileSerializer(serializers.ModelSerializer):
...
 
-    download_url = serializers.CharField(source='private_download_url')
     metadata = serializers.JSONField()
 
     class Meta:  # noqa: D101
         model = DataFile
         fields = ('id', 'basename', 'created', 'download_url', 'metadata',
                   'source')
+
+    def to_representation(self, instance):
```
madprime: Is it possible to...
1. override `_readable_fields` to `super()` then remove `download_url` from the returned set
2. super() `to_representation`, then add `download_url`
?

Thinking about how to do this in a way that minimizes customization on REST framework :)

data_import/views.py
```diff
@@ -46,12 +71,9 @@ def get(self, request, *args, **kwargs):
...
             return self.get_and_log(request)
 
-        query_key = request.GET.get('key', None)
-        if query_key:
-            key_qs = DataFileKey.objects.filter(datafile=self.data_file)
-            key_qs = key_qs.filter(key=query_key)
-            if key_qs.exists():
-                if not key_qs.get().expired:
-                    return self.get_and_log(request)
+        key = self.get_key(request)
```
madprime: redundant?

data_import/models.py
jhdulaney: Yeah, I grepped for it and only found it in two places, here in the model and in the serializer.  I'm certain it's safe to remove.

data_import/serializers.py
jhdulaney: The docs do specifically state:
There are four methods that can be overridden, depending on what functionality you want the serializer class to support:

    .to_representation() - Override this to support serialization, for read op

So, doing it this way seems to be more in line with the way that rest-framework wants things done?

data_import/views.py
jhdulaney: oh, lol

data_import/serializers.py
madprime: But this code seems like it's a copy/paste of some version of `ModelSerializer`... but not the current `master`. And it's relying on `_to_representation` having a particular behavior, so... I don't think this makes less assumptions than my suggestion.

Anything that is copy/paste, and especially when the source it lifted from is changing, troubles me -- my preference to actually defer to whatever REST framework is doing, with a minimal override, not copy/paste it.

madprime: Obsoleting this; the function is rewritten from scratch now :+1:

Issue #944, opened 2019-01-07
Accessing the API with a master token for onsite projects causes an attribute error


jhdulaney:
When accessing /api/direct-sharing/project/members/ with an access_token, request.auth gets set directly to the project object rather than to an auth object which contains the project object as a parameter.

jhdulaney:
Fixed #945

Issue #943, opened 2019-01-04
Race condtion between datafile deletion and datafilekey creation


jhdulaney:
There is a race condtion where a datafile looks to be deleted but a key for it is created, causing a foreign key error.

fix would be:

"if I did something like logging datafile info with the key rather than having a foreignkey relation in the db
that'd a) clear up this race condition and b) save the information"

jhdulaney:
Fixed #945

PR #945, opened 2019-01-07
Transactions 20190107

jhdulaney:
## Description
Stores datafile id rather than foreign key
Checks for datafile existence before attempting to retrieve it in case file gets deleted during datafile key's life span
accounts for different auth method for onsite projects

## Related Issue
#944 and #943

## Testing
Reproduced #944
Tested /api/direct-sharing/project/members/?access_token=foo
tested member exchange endpoint
Tested download from activity page
Reproduced #943 
Retested member exchange endpoint by bringing up the page, deleting data, then trying to access the datafile.


data_import/views.py
```diff
@@ -47,11 +47,14 @@ def get_and_log(self, request, key_object=None):
...
-        self.data_file = DataFile.objects.get(pk=self.kwargs.get('pk'))
-
-        unavailable = (
-            hasattr(self.data_file, 'parent_project_data_file') and
+        data_file_qs = DataFile.objects.filter(pk=self.kwargs.get('pk'))
+        if data_file_qs.exists():
+            self.data_file = data_file_qs.get()
+            unavailable = (
+                hasattr(self.data_file, 'parent_project_data_file') and
             self.data_file.parent_project_data_file.completed is False)
```
madprime: Indent this too?

jhdulaney: oh, woops

Issue #942, opened 2019-01-04
Improve main page load time


madprime:
The main page (`/`) in production seems to have slow load times, with a median around 1.2 to 1.3 seconds.

jhdulaney:
Tried a couple of different page speed analyzers, and our biggest issue is images.  Image load times seem to range from 1 to 4 seconds.

jhdulaney:
Also, Google recommends sending js in smaller payloads.

jhdulaney:
Fixed #946.

PR #946, opened 2019-01-10
Main site optimizations

jhdulaney:
## Description
Prior to this change, we were hitting the database more times than were necessary, but also we were iterating over the database results multiple time.  This caused data display to take quadratic time.

This changes to using single database queries and then only iterating over the results once to actually serialize the data for display.

This applies to the following endpoints:
/
/create/
/members/
/member/me/joined/

As an example of the difference this change causes, django-silk reported a load time of 1314ms overall and 515ms for the 78 db queries in order to render the home page.  With this change, that becomes 683ms overall with 103 ms spent on the 14 db queries.  This difference is consistent and reproducible.

## Related Issue
#942 

## Testing
Largely consisted of comparing running locally vs the production site with a local copy of the production database.
Automated tests pass
Using django-silk to compare load times


open_humans/views.py
```diff
@@ -287,14 +288,20 @@ def get_context_data(self, *args, **kwargs):
...
-            approved=True).filter(
-                active=True)
-        activities = sort_activities(
-            {proj.id_label: activity_from_data_request_project(
-                project=proj, user=self.request.user) for proj in projects})
+            approved=True).filter(active=True).exclude(
+                returned_data_description__isnull=True).exclude(
+                    returned_data_description__exact='')
+        if not self.request.user.is_anonymous:
+            project_memberships = DataRequestProjectMember.objects.filter(
```
madprime: When I test locally it seems that this filter isn't working - I log in with my user and see projects listed that I've joined...

Also, previously this list was sorted according to number of members - I think we should keep doing that :)

madprime: Also: ExploreSharePageView was re-using this view with another template.

But because the filtering on "data source" was moved from template to view (probably appropriate), the ExploreSharePageView is now broken -- it's producing the same list.

open_humans/views.py
```diff
@@ -41,6 +39,17 @@
 TEN_MINUTES = 60 * 10
 
 
+def sort_projects_by_membership(projects):
+    """
+    Takes a queryset of projects and returns a queryset sorted by the number of
+    members in a project.
+    """
+    projects = projects.annotate(num_members=Count(
+        'project_members', filter=Q(project_members__joined=True)))
```
madprime: Oops, I've realized that unfortunately this is still counting suspended users -- RunKeeper in particular had a fiasco that produced a large batch of suspended users. I think it'll need to add something like this...

`.annotate(num_members=Count('project_members', filter=Q(project_members__joined=True) & Q(project_members__member__user__is_active=True)))`

private_sharing/forms.py
```diff
@@ -73,6 +73,30 @@ def __init__(self, *args, **kwargs):
...
+        Logic to for conditional required elements in our form.
+        """
+        cleaned_data = super().clean()
+
+        add_data = cleaned_data.get('add_data', False)
+        explore_share = cleaned_data.get('explore_share', False)
+        returned_data_description = cleaned_data.get(
+            'returned_data_description', None)
+
+        if not add_data or explore_share:
```
madprime: I think you mean: `if not (add_data or explore_share):`

private_sharing/forms.py
```diff
@@ -73,6 +73,30 @@ def __init__(self, *args, **kwargs):
...
+            self.add_error('add_data',
+                           forms.ValidationError('Please pick "Add data" or '
```
gedankenstuecke: I think this error message isn't fully correct. People can assign their projects to be both, 'add data' and 'explore & share'. So it should be re-written to 'Pick at least one option from 'add data' and 'explore & share'" or something along the lines.

jhdulaney: Done

PR #961, opened 2019-01-23
Add email notification to suspend_users command

madprime:
Adds an email notification to the account suspension management command.

Tested locally with console email and local database.

### DIFF
--- a/open_humans/management/commands/suspend_users.py
+++ b/open_humans/management/commands/suspend_users.py
@@ -1,4 +1,5 @@
 from django.contrib.auth import get_user_model
+from django.core.mail import get_connection, EmailMultiAlternatives
 from django.core.management.base import BaseCommand, CommandError
 
 UserModel = get_user_model()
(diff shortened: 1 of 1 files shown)


open_humans/management/commands/suspend_users.py
```diff
@@ -44,7 +53,40 @@ def handle(self, *args, **options):
                     raise CommandError('User ID "{}" does not exist!'
                                        .format(id_str))
 
+        if options['email']:
+            self.email_notification(users_to_suspend)
```
gedankenstuecke: Do I see this correctly that the email notifications about suspension go out even if `skip_suspend` is active and no actual suspension is performed?

madprime: If `--email --skip-suspend` is used then yes.

I added this because I was concerned about a scenario where suspension had been done but the `--email` flag was forgotten, in which case this would be the correction... but... maybe that makes no sense, it's harmless to go ahead with suspend?

gedankenstuecke: Ah, sorry. I had missed the email flag, making that optional too, so one can recover from both mistakes (sending email, not suspending & vice versa). Great! 🙂

madprime: nah, on reflection your question makes it clear that "skip-suspend" is unnecessary, it's harmless to "re-suspend" users to redo the command to send emails. so I've removed it, thanks!

gedankenstuecke: ok! 👍

Issue #954, opened 2019-01-18
Add a field to the ProjectMember model that stores when a member last joined a given project


jhdulaney:
## Description
Add a field to the ProjectMember model that stores when a member last joined a given project

## Related Issue(s)
#926

jhdulaney:
Merged #962

PR #962, opened 2019-01-24
Record when a user joins and authorizes a project

jhdulaney:
## Description
This change records when a user joins a project, and also records when a user authorizes a project.

## Related Issue
#954 

## Testing
Made a new test user
Joined project, ensured that the database was updated appropriately
Authed project, checked db
Left and deauthed project, ensured database last_joined and last_authorized fields did not change
rejoined project, ensured database was updated.


private_sharing/models.py
```diff
@@ -526,10 +538,21 @@ def leave_project(self, remove_datafiles=False, done_by=None, erasure_requested=
             items.delete()
 
     def save(self, *args, **kwargs):
+        """
+        Oviride to record when a project is joined and authorized and generate a
```
madprime: Typo here. :)

Also a general question -- maybe @gedankenstuecke has an opinion too -- would it be better to be recording all join/authorization history, as was done for the record of project approval (#948)?

gedankenstuecke: Mh... I think it wouldn't harm to store that data instead of losing it (even if I don't have a immediate use case in mind)!

jhdulaney: Went ahead and did that thing

PR #968, opened 2019-01-31
Reduce number of database hits for exchange-member endpoint

jhdulaney:
## Description
This change reduces the number of database hits on the exchange member endpoint, from over a thousand for a large set of files, to under 20.

## Related Issue
Server slowness

## Testing
Was able to reproduce on a local copy of the production database using an access_key for a user that was known to have a large file set.
Silk profiling indicates greatly reduced number of queries, plus greatly reduced execution time

Tested other endpoints that hit the same code in the datafile model to ensure that they continued to function as expected.

### DIFF
--- a/data_import/models.py
+++ b/data_import/models.py
@@ -125,7 +125,10 @@ class DataFile(models.Model):
         return self.file.storage.url(self.file.name)
 
     def private_download_url(self, request):
-        if self.is_public:
+        if hasattr(request, 'public_sources'):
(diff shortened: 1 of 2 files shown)


data_import/models.py
```diff
@@ -150,7 +153,7 @@ def generate_key(self, request):
         return new_key.key
 
     @property
-    def is_public(self):
+    def is_public(self, public_sources=None):
```
madprime: what's `public_sources` for here?

PR #976, opened 2019-02-12
Update public data badge and default profile pic

### LINKED ISSUE #814: Update Public Data badge
## General Checkups

## Description

We forgot to update the `public data sharing` badge when doing the redesign (probably as it's not a proper project 😂). Should also get a makeover with the new color scheme.

madprime:
Fixes #814


jhdulaney:
Looks better to me

PR #986, opened 2019-02-21
WIP Initial datafile datatypes

jhdulaney:
## Testing


private_sharing/models.py
```diff
@@ -631,6 +632,15 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
 
 
+class ProjectOntology(models.Model):
```
madprime: Why isn't this a new ManyToManyField on the project model itself?

(Also, I don't see how this model connects to a project as it's currently written...)

private_sharing/models.py
```diff
@@ -655,7 +665,7 @@ class ProjectDataFile(DataFile):
         related_name="parent_project_data_file",
         on_delete=models.CASCADE,
     )
-
+    categories = models.ManyToManyField(Ontology, related_name="datafile_categories")
```
madprime: Regarding "categories" and "Ontology": I think "DataType" everywhere would be more consistent and clear.

- the model "Ontology" could be "DataType"
- the field "categories" could be "datatypes"

private_sharing/models.py
```diff
@@ -236,6 +236,7 @@ class DataRequestProject(models.Model):
         ),
         verbose_name="Are you requesting Open Humans usernames?",
     )
+    data_types = models.ManyToManyField("ProjectOntology", related_name="data_types")
```
madprime: it's unclear why you'd want a manytomany to ProjectOntology which, in turn, has manytomany to Ontology. AFAICT there's just one target: DataType, and each project may have zero, one, or more DataTypes it's registered.

This field is ambiguously named - is it the datatypes the project is registering for allowed uploads? Or is it replacing requested_sources for the requested data access?

private_sharing/api_views.py
```diff
@@ -226,7 +227,31 @@ def post(self, request, format=None):
...
+class SaveDataTypesMixin(object):
+    """
+    Mixin to do some final verification of datatypes to be associated with the datafile
+    and to do the actual association.
+    """
+
+    @property
+    def good_datatypes(self):
+        """
+        Verifies that the supplied datatypes are valid.  Returns boolean.
```
madprime: Was there some reason this couldn't be done in the form?

private_sharing/api_views.py
```diff
@@ -301,6 +330,10 @@ class ProjectFileUploadView(ProjectFormBaseView):
     def post(self, request):
         super().post(request)
 
+        # Check datatypes
+        if not self.good_datatypes:
```
madprime: (this feels out of place since normally it'd be raised as an error in the form validation)

private_sharing/api_views.py
```diff
@@ -226,7 +227,31 @@ def post(self, request, format=None):
...
+        return self.form.cleaned_data["datatypes"].issubset(datatypes_ids)
+
+    def save_datatypes(self, data_file):
+        """
+        Saves the provided datatypes in the ProjectDataFile instance.
+        """
+        for datatype_id in self.form.cleaned_data["datatypes"]:
+            datatype = self.project.datatypes.get(id=datatype_id)
+            data_file.datatypes.add(datatype)
+        data_file.save()
```
madprime: Can you just add the datatypes, then make sure to save the model after calling this function (not before)?

private_sharing/api_views.py
```diff
@@ -247,6 +275,7 @@ def post(self, request):
         )
 
         data_file.save()
+        self.save_datatypes(data_file)
```
madprime: See above...

private_sharing/forms.py
```diff
@@ -397,3 +419,32 @@ class DeleteDataFileForm(forms.Form):
...
+    """
+
+    def clean(self):
+        cleaned_data = super().clean()
+        # Check to see if anything was selected
+        # wants to be on two lines:
+        data = dict(self.data)
+        data.pop("csrfmiddlewaretoken")
+        if data:
+            cleaned_data.update(self.data)
```
madprime: this doesn't seem to be used?

private_sharing/forms.py
```diff
@@ -397,3 +419,32 @@ class DeleteDataFileForm(forms.Form):
...
+            self.cleaned_data = data
```
madprime: is this overriding/redoing stuff that super() already does?

Maybe this method could be reduced to simply raise a `ValidationError` if at least one hasn't been selected? Otherwise, return the super()?

private_sharing/models.py
```diff
@@ -236,6 +236,7 @@ class DataRequestProject(models.Model):
         ),
         verbose_name="Are you requesting Open Humans usernames?",
     )
+    datatypes = models.ManyToManyField(DataTypes, related_name="data_types")
```
madprime: I think it's unnecessary to have a related_name here, the default of `datatype_set` nomenclature seems fine to me (I like sticking to defaults as being more predictable) :)

private_sharing/models.py
```diff
@@ -655,7 +656,7 @@ class ProjectDataFile(DataFile):
         related_name="parent_project_data_file",
         on_delete=models.CASCADE,
     )
-
+    datatypes = models.ManyToManyField(DataTypes, related_name="datafile_datatypes")
```
madprime: how about `registered_datatype`? then you can skip the related_name, I think the default would be `project.registered_datatype_set`

Note... there isn't any `requested_datatype` yet but something like that will also exist (although notably sometimes the datatypes will have project-specificity, implementation TBD...)

data_import/models.py
```diff
@@ -224,3 +224,39 @@ class TestUserData(models.Model):
...
+    """
+
+    name = models.CharField(max_length=128, blank=False, unique=True)
+    parent = models.ForeignKey("self", blank=True, null=True, on_delete=models.CASCADE)
+    description = models.CharField(max_length=512, blank=False)
+    created = models.DateTimeField(auto_now=True)
+
+    def __str__(self):
+        parents = self.get_all_parents
+        if parents:
```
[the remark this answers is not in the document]
jhdulaney: Except parents is a list and I do things with it?

private_sharing/api_views.py
jhdulaney: Because passing objects to the form requires much more complexity and this is three lines plus a doc string.

private_sharing/api_views.py
jhdulaney: See above about having to pass objects to the form

private_sharing/api_views.py
jhdulaney: The documentation explicitly states that you cannot associate objects with a manytomany field until the object has been saved; it must be done separately.

private_sharing/forms.py
jhdulaney: By this point, the csrf token has been checked and is no longer necessary, and removing it makes logic later on easier.

private_sharing/forms.py
jhdulaney: cleaned_data.update(self.data) is really not necessary here; otherwise returning the dict without the csrf token does make processing later easier.

the super() does not actually populate cleaned_data if there are no fields formally defined, making this override necessary.

jhdulaney:
Closing to be reopened against a feature branch.

PR #987, opened 2019-02-21
[WIP] Bootstrap 4 & new activity page

madprime:
Currently "in progress"...
 - new base template for bootstrap 4
 - new SCSS and CSS files for bs4 version
 - new login and signup partials for bs4 version
 - initial work on an activity page
 - a test view for checking global styles

Note: pages that use the new Bootstrap 4 based templates no longer use the old CSS files.


open_humans/templates/account/login-social-bs4.html
```diff
@@ -0,0 +1,30 @@
+{% load socialaccount %}
+{% load static %}
+{% load utilities %}
+
+<link href="https://fonts.googleapis.com/css?family=Roboto" rel="stylesheet">
```
jhdulaney: I added Roboto to the static site.

Issue #981, opened 2019-02-13
Add DataTypes


madprime:
## Description
Currently management of data authorizations is performed according to "data source" - that is, the project that produced the data. This is limiting! We believe some representation of "type" of data, and enabling members and projects to manage permissions according to this, would improve usability of the platform.

### Some examples of why we need "data types"

1. **Granular permission** Project A generates survey + sensitive data (e.g. genetic) and Project B just wants access to one "type", not both.

2. **Shared permission** Project A generates data of type X. Project B also generates data of type X. Project C would like to get data of type X -- but the only way to do this currently is to ask for each project (Project A and Project B).

3. **Future-friendly authorization** Project D adds data of type X. In the current model, Project C now needs to add "Project D" to its list of authorizations. Furthermore, members need to re-authorize Project C to grant that access.

## Specs

We imagine that a "DataType" could exist that has the following features:
1. unique label/identifier
2. description
3. parent DataType (can be null)

We won't support multiple inheritance.

Furthermore, the project itself is a final characterization of a data type. We might think of these as terminal nodes in some tree ontology, e.g. "genomic data:23andMe data:data from direct-sharing-128". But it might not be necessary to do more than consider an optional specification of "project source" alongside "data type" in authorizations.

### Data source behavior

1. **Registration of data types** Projects should register DataTypes during configuration; these will be reviewed as part of the approval process.
2. **API enforcement** The API should require file uploads to have at least one DataType. They may have more than one. All DataTypes must match ones that were registered by the project.
3. **Re-approval** Changing registration of DataTypes requires re-review.

The rational for requiring re-approval is to reduce the chances of a project mistakenly using DataTypes in a manner which causes undesired data sharing (e.g. mislabeling GPS data as survey data, rendering it available to a project that requested survey data).

### Authorization behavior

Projects may request one or more DataTypes. Each DataType request may optionally be specific to a certain project source. For example:

* genetic data (any source)
* survey data -- from "Harvard PGP" source

Project source is optional. Unlike the current behavior, authorization will never be according to project source alone.

Data with multiple types assigned is authorized if any of those types is requested.

As is currently done, Member authorization is for what was requested at the time of joining a project. Re-authorization is needed for new requests.

jhdulaney:
So, I'm thinking
we have a tree
it collapses down to some number of base types
this lowest level is editable only by OH staff
you click one, it expands
and, so on, through to the leaves
Each node after these base types gets a checkbox
[...]

PR #993, opened 2019-02-22
Datafile Datatypes

jhdulaney:
## Description
This pull request adds the concept of Datatypes to Open Humans.  It provides a mechanism for site users to add new datatypes to the project.  This mechanism is available to all Open Humans members, though it is only linked to the from the Select Datatypes View, which in turn is only available to Project Coordinators.  In turn, projects immediately after creation are sent to this Select Datatypes View, where they can choose the datatypes they will be using with their project, as well as add new datatypes.

This change also enforces that data_files must have one or more datatypes.

## Related Issue
#981 

## Testing
Added to the automated testing

Manually tested:
Create a project, selected a set of datatypes, added a new datatype
Checked the database directly to ensure datatypes were actually saved
Attempted file upload without specifying a datatype, noted that it failed
Attempted file upload with incorrect datatypes, noted it failed
File upload with full set of project datatypes, succeeded
File upload with subset of project datatypes also succeeded.


private_sharing/api_views.py
```diff
@@ -301,6 +344,10 @@ class ProjectFileUploadView(ProjectFormBaseView):
     def post(self, request):
         super().post(request)
 
+        # Check datatypes
+        if not self.good_datatypes:
```
madprime: This seems like a form validation step -- not something that goes in the view.

private_sharing/views.py
```diff
@@ -676,3 +686,123 @@ def get_object(self, queryset=None):
...
+    def form_valid(self, form):
+        """
+        Form is good, let's save this thing.
+        """
+        ret = super().form_valid(form)
+        if "add-datatype" in form.cleaned_data:
+            # User wants to add a new datatype.  Save the current state of the
+            # selection form.
+            form.cleaned_data.pop("add-datatype")
+            self.request.session["select_category_form"] = form.cleaned_data
```
madprime: it strikes me as very unusual to be using session when filling in forms, I think I'd like to understand why it's being done this way (assuming you're confident this is the correct thing to do)

private_sharing/forms.py
```diff
@@ -393,3 +420,35 @@ class DeleteDataFileForm(forms.Form):
...
+    def clean(self):
+        """
+        Check that something was actually passed, and, if so, populate and return
+        cleaned_data.
+        """
+        super().clean()
+        # Check to see if anything was selected
+        # wants to be on two lines:
+        data = dict(self.data)
+        data.pop("csrfmiddlewaretoken")
```
madprime: This jumps out to me as something weird/wrong to have to do. I believe normally custom clean functions use `cleaned_data` e.g. for complex validation: https://docs.djangoproject.com/en/2.1/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other

private_sharing/views.py
jhdulaney: I'm just using the session to temporarily save form state so that when we return to the form we can reload it.

PR #996, opened 2019-03-06
More datatypes work

jhdulaney:
## Description
Adds the auto_add_datatypes field for grandfathering in existing projects, adds a new API endpoint for displaying datatypes, and does some minor tweaks.

## Related Issue
#981 

## Testing
Set the auto_add_datatypes boolean via admin interface
Uploaded a datafile without datatype specified.
Verified that the datafile had the full set of project datatypes associated with it.


private_sharing/api_views.py
```diff
@@ -254,6 +264,9 @@ def save_datatypes(self, data_file):
 
         datatypes can be looked up either via name or ID
         """
+        if self.project.auto_add_datatypes:
```
madprime: I think we should do this only if datatypes weren't specified.

private_sharing/serializers.py
```diff
@@ -116,3 +116,23 @@ def to_representation(self, obj):
             rep.pop("username")
 
         return rep
+
+
+class DataTypeSerializer(serializers.ModelSerializer):
+    """
+    Serialize data for a project member.
```
madprime: incorrect due to copy/paste?

private_sharing/serializers.py
```diff
@@ -116,3 +116,23 @@ def to_representation(self, obj):
...
+    class Meta:  # noqa: D101
+        model = DataType
+
+        fields = ["id", "name", "parent", "description", "projects"]
+
+    projects = serializers.SerializerMethodField()
+
+    def get_projects(self, obj):
+        """
+        Get the other sources this project requests access to.
```
madprime: incorrect due to copy/paste?

private_sharing/serializers.py
```diff
@@ -116,3 +116,23 @@ def to_representation(self, obj):
...
+    """
+
+    class Meta:  # noqa: D101
+        model = DataType
+
+        fields = ["id", "name", "parent", "description", "projects"]
+
+    projects = serializers.SerializerMethodField()
+
+    def get_projects(self, obj):
```
madprime: This needs to distinguish between source_projects and requesting_projects. (It's just that so far only source has been implemented, but the intent is to include requesting ones as well.)

private_sharing/api_views.py
```diff
@@ -428,3 +441,23 @@ def post(self, request):
...
+    Lists the datatypes available and which projects use them.
+    """
+
+    serializer_class = DataTypeSerializer
+
+    def get_queryset(self):
+        """
+        Get the queryset and filter on project if provided.
+        """
+        project_label = self.request.GET.get("project", None)
```
madprime: should distinguish between source project and requesting project (this is source)

private_sharing/api_views.py
jhdulaney: Roger

private_sharing/serializers.py
jhdulaney: Roger that

PR #999, opened 2019-03-08
Add case insensitive unquieness verification to new datatypes

jhdulaney:
## Description
This change verifies that new datatype names are unique despite case

## Related Issue
N/A

## Testing
Created datatype
Attempted to create a new datatype except all upper case
Verified that the form rejected the new datatype.


private_sharing/api_urls.py
```diff
@@ -20,6 +20,7 @@
         "project/files/upload/complete/",
         api_views.ProjectFileDirectUploadCompletionView.as_view(),
     ),
+    path("datatypes/", api_views.ListDataTypesView.as_view()),
```
madprime: Because the DataTypes model is defined in the `data_import` app, I'm concerned that it's inappropriate to be placing DataType-specific code in other apps. I don't hold this as a blocker for this PR, but I think it should be resolved before merging.

In this case, because there's already "public data" API endpoints in the `open_humans` app, I suspect that's the correct place for this (as another publicly available endpoint).

PR #1000, opened 2019-03-09
[WIP] 201902 feature datafile datatypes

madprime:
PR opened for commenting, not ready to merge.

This feature breanch is intended to partly implement #981:
* add models and views for creating DataTypes
* add views for managing a project's registered DataTypes
* handle specified DataTypes in the API for file uploads
* some default behavior if DataTypes weren't specified during upload


private_sharing/forms.py
```diff
@@ -393,3 +424,37 @@ class DeleteDataFileForm(forms.Form):
     file_basename = forms.CharField(required=False, label="File basename")
 
     all_files = forms.BooleanField(required=False, label="All files")
+
+
+class SelectDatatypesForm(forms.Form):
```
madprime: I think this form and associated view are more cleanly done with a ModelForm and UpdateView.

data_import/models.py
```diff
@@ -224,3 +230,41 @@ class TestUserData(models.Model):
...
+
+class DataType(models.Model):
+    """
+    Describes the types of data a DataFile can contain.
+    """
+
+    name = models.CharField(
+        max_length=128, blank=False, unique=True, validators=[charvalidator]
+    )
+    parent = models.ForeignKey("self", blank=True, null=True, on_delete=models.CASCADE)
```
madprime: here's a good spot to use `related_name=children` so one can do `parent_datatype.children.all()`, otherwise it's `parent_datatype.datatype_set.all()` which isn't as intuitive.

also I think `on_delete=models.CASCADE` is a really bad idea for DataTypes. I would go with `PROTECT` to prevent orphaned DataTypes in the tree.

private_sharing/api_views.py
```diff
@@ -226,7 +233,50 @@ def post(self, request, format=None):
...
+    def good_datatypes(self):
+        """
+        Verifies that the supplied datatypes are valid.  Returns boolean.
+        """
+        # running this test here as we have the object available in the view, but
+        # not in the form.
+        # First, we create a set containing all possible IDs and names for a project's
+        # datatypes.  We then check that the requested datatypes are a subset, which,
+        # in Python, can include any portion of the set up to the entire set.
+        if self.project.auto_add_datatypes:
```
madprime: This should only be done if datatypes weren't specified. It's a fallback.

private_sharing/forms.py
```diff
@@ -313,6 +315,35 @@ class UploadDataFileBaseForm(forms.Form):
...
+        # Note:  To be removed when strict datatypes are required
+        if not datatypes:
+            return set()
+        if not (datatypes.startswith("[") and datatypes.endswith("]")):
+            raise forms.ValidationError(
+                "A list of datatypes is required to describe this file"
+            )
+        # Sort the requested IDs and turn them into ints so that we match what will
+        # come out of the database
+        datatypes_list = ast.literal_eval(datatypes)
```
madprime: It seems like it would be safer to use a data serialization format like JSON? (I didn't realize this was even possible - so that's interesting to learn, but...)

private_sharing/api_views.py
```diff
@@ -226,7 +233,50 @@ def post(self, request, format=None):
...
+    and to do the actual association.
+    """
+
+    @property
+    def good_datatypes(self):
+        """
+        Verifies that the supplied datatypes are valid.  Returns boolean.
+        """
+        # running this test here as we have the object available in the view, but
+        # not in the form.
```
madprime: It _is_ possible to pass something to a form when instantiated, for example `BaseProjectMembersForm`'s `__init__` method setting a `self.project`: https://github.com/OpenHumans/open-humans/blob/master/private_sharing/forms.py#L158

The generic FormView makes this customizable with a `get_form_kwargs` method. It looks like `ProjectFormBaseView` doesn't provide something like this, so... not worth worrying about.

private_sharing/api_views.py
```diff
@@ -235,7 +285,10 @@ class ProjectFileDirectUploadView(ProjectFormBaseView):
     form_class = DirectUploadDataFileForm
 
     def post(self, request):
-        super(ProjectFileDirectUploadView, self).post(request)
+        super().post(request)
+
+        if not self.good_datatypes:
+            return HttpResponseForbidden()  # Did not include a properly formed datatype
```
madprime: This should be resulting in a serializer error, as with other API views, e.g.:
https://github.com/OpenHumans/open-humans/blob/master/private_sharing/api_views.py#L160

private_sharing/api_views.py
```diff
@@ -270,11 +324,14 @@ class ProjectFileDirectUploadCompletionView(ProjectFormBaseView):
...
-        super(ProjectFileDirectUploadCompletionView, self).post(request)
-
-        data_file = ProjectDataFile.all_objects.get(
-            pk=self.form.cleaned_data["file_id"]
-        )
+        super().post(request)
+        # If upload failed, file_id would be empty.  Let's fail gracefully.
+        file_id = self.form.cleaned_data.get("file_id", None)
+        if not file_id:
+            return Response(
```
madprime: Seems like this hsould be raising an error the APIView handles?
e.g. https://github.com/OpenHumans/open-humans/blob/master/private_sharing/api_views.py#L160

private_sharing/api_views.py
```diff
@@ -301,6 +358,10 @@ class ProjectFileUploadView(ProjectFormBaseView):
     def post(self, request):
         super().post(request)
 
+        # Check datatypes
+        if not self.good_datatypes:
+            return HttpResponseForbidden()  # Did not include a properly formed datatype
```
madprime: raise serializers.Validation error

madprime:
obsolete, see #1026

Issue #782, opened 2018-05-16
Finalize legacy app removal


madprime:
Items left undone in legacy project conversions:

- [ ] Remove original data files for apps that received and processed raw data uploads
- [ ] Remove models for apps and perform migration
- [ ] Remove apps themselves
- [ ] Remove `LEGACY_APPS` filter in `common/utils.py`

jhdulaney:
Going to use this as the tracking issue for upcoming site cleanup work.

jhdulaney:
https://github.com/OpenHumans/open-humans/blob/009c05cac9f8755412e33bfef872d01667749ef5/private_sharing/forms.py#L224

This loops over all members for sending messages, needs to be cleaned up.

PR #1007, opened 2019-03-12
Remove dead code from common.activities

jhdulaney:
## Description
This commit removes obsolete and dead code from common.activities.  Note that the final removal of this module is dependent on the new activities page.

## Related Issue
#782 

## Testing
Automated tests pass
General functionality tests:
- Created project
- Checked that members had badges in /members


open_humans/templatetags/utilities.py
```diff
@@ -217,19 +216,6 @@ def active(context, pattern_or_urlname):
     return ""
 
 
-@register.simple_tag()
-def url_slug(label):
```
madprime: I guess the `source_to_url_slug` is also no longer needed... (not a blocker though -- a random spot check, now I wonder what other templatetags exist that are obsolete)

jhdulaney: Yeah, I'll have a look at those at some point.

PR #1011, opened 2019-03-20
Enable retrieving AWS access logs

jhdulaney:
## Description
Retrieves AWS log files from the log file bucket and populates a new database model, AWSDataFileAccessLog with the log entries.

This is meant to be run as a cron job; the more often it gets run, the lighter the load would be.  I suggest every ten minutes.

This change adds the AWSDataFileAccessLog model as a distinct model from NewDataFileAccessLog because, while the former logs access on Amazon's end, the latter logs access from Open Humans end, and these can be distinct (though related) events.  Wherever possible, the relation between the two is recorded, though it is not necessarily guaranteed that such a relation will exist.

The data from the log files is parsed out and used to populate the database.  Wherever possible, conversion to the correct datatypes is performed (eg, int and datetime objects).  Population of the individual fields is done as per the ordering documented by Amazon [0].  The documentation states that the log format may be extended, but implies that the current ordering would be kept the same.

Once a log file has been retrieved and parsed, it gets deleted from the bucket.

There are two things that this change does not do that should be discussed.

The first is enabling file access keys for all files on open humans, rather than just those privately shared.  This was not done previously due to worries over performance.  However, the performance bottleneck has been resolved via pagination of API endpoints, so this is no longer a consideration.

The second is rotating logs.  Provision is made for this via the 'created' field in the model.  This could be achieved via:
```AWSDataFileAccessLog.objects.filter(created__lte=X_DAYS).delete()``` or similar, run as a daily cron job.  This was not implemented because I would like to discuss whether or not we actually want to do this.

I note that if it is decided to implement either of these two changes, it would be a rather quick job and could be the subject of separate changes.

One final note is that, due to the changes to the NewDataFileAccessLog model, a one off migration command will be necessary, and then a new change to remove the existing data_file_key field, which will no longer be needed.

[0]  https://docs.aws.amazon.com/AmazonS3/latest/dev/LogFormat.html

## Related Issue
N/A, discussed during meeting.

## Testing
Created a dev log bucket
Tested against empty log bucket
Ran against full log bucket, ensured that database was correctly populated.

### CODE REVIEW

File: open_humans/management/commands/vacuum_log_bucket.py [hunk 1]
```diff
@@ -0,0 +1,99 @@
...
+        self.stdout.write("Retreiving file access logs")
+        s3 = boto3.resource("s3")
+        bucket = s3.Bucket(settings.LOG_BUCKET)
+        for item in bucket.objects.all():
+            log_objects = item.get()["Body"].read().decode("utf-8").split("\n")
+            for log_entry in log_objects:
+                if not log_entry:
+                    continue
+                log = []
+                start_chars = ['"', "["]
```
madprime: It seems unlikely that you should need to write a new "quote-protected split" yourself. I googled `quote-protected split python` and found this?

https://stackoverflow.com/questions/79968/split-a-string-by-spaces-preserving-quoted-substrings-in-python

open_humans/management/commands/vacuum_log_bucket.py
```diff
@@ -0,0 +1,99 @@
...
+                            tmp_index = len(log)
+                    if (not tmp_index) or (tmp_index == len(log)):
+                        log.append(string)
+                        continue
+                    log[tmp_index] = "{0} {1}".format(log[tmp_index], string)
+                    if string[-1] in end_chars:
+                        tmp_index = 0
+
+                aws_log_entry = AWSDataFileAccessLog()
+                aws_log_entry.bucket_owner = log[0]
```
madprime: This set of code could be redone?
1. an array ['bucket_owner', 'bucket', ...]
2. for loop over range(len(array))
3. setattr(aws_log_entry, array[i], log[i])

data_import/views.py
```diff
@@ -26,27 +25,22 @@ def get_and_log(self, request, key_object=None):
...
+            access_log.project_id = key_object.project_id
+            access_log.access_token = key_object.access_token
+            access_log.key_creation_ip_address = key_object.ip_address
+            access_log.key_created = key_object.created
+            access_log.key = key_object.key
+            url = "{0}&x-oh-key={1}".format(
+                self.data_file.file_url_as_attachment, key_object.key
+            )
+        else:
+            url = self.data_file.file_url_as_attachment
```
madprime: I'm noticing that we don't log access events by the user themselves. (Is that correct?) I think ... maybe we should be doing that, for a variety of reasons? Evidence/records if a "leak" occurs because of unauthorized access to a user's account. What do you think?

open_humans/management/commands/vacuum_log_bucket.py
jhdulaney: Ahhh, your Google-foo is better than mine!

data_import/views.py
jhdulaney: Yeah, that's why I was thinking about enabling key access across the board.

open_humans/management/commands/vacuum_log_bucket.py
```diff
@@ -0,0 +1,82 @@
...
+        self.stdout.write("Retreiving file access logs")
+        s3 = boto3.resource("s3")
+        bucket = s3.Bucket(settings.LOG_BUCKET)
+        for item in bucket.objects.all():
+            log_objects = item.get()["Body"].read().decode("utf-8").split("\n")
+            for log_entry in log_objects:
+                if not log_entry:
+                    continue
+                parser = ZeroOrMore(
+                    Regex(r"\[[^]]*\]") | Regex(r'"[^"]*"') | Regex(r"[^ ]+")
```
madprime: Why use a regex here instead of `shlex` (which was described in the stackoverflow link I shared before)?

```
>>> import shlex
>>> teststr = 'this is a "string of words" that has some "in quotes" together'
>>> shlex.split(teststr)
['this', 'is', 'a', 'string of words', 'that', 'has', 'some', 'in quotes', 'together']
```

open_humans/management/commands/vacuum_log_bucket.py
```diff
@@ -0,0 +1,82 @@
...
+                )
+                log = list(parser.parseString(log_entry))
+
+                aws_log_entry = AWSDataFileAccessLog()
+                skip_fields = ["id", "created", "data_file", "oh_data_file_access_log"]
+                fields = [
+                    field
+                    for field in aws_log_entry._meta.get_fields()
```
madprime: I'm concerned that this is putting a lot of trust in the ordering of get_fields remaining the same, I don't think that's a good idea.

open_humans/management/commands/vacuum_log_bucket.py
```diff
@@ -0,0 +1,82 @@
...
+                oh_key = parse_qs(parsed_url.query).get("x-oh-key", [""])[0]
+                if oh_key == "None":
+                    oh_key = None
+                if oh_key:
+                    oh_data_file_access_logs = NewDataFileAccessLog.objects.filter(
+                        key=oh_key
+                    )
+                    aws_log_entry.oh_data_file_access_log.set(oh_data_file_access_logs)
+
+            # item.delete()
```
madprime: ?

open_humans/settings.py
```diff
@@ -416,6 +416,8 @@ def to_bool(env, default="false"):
 AWS_SECRET_ACCESS_KEY = os.getenv("AWS_SECRET_ACCESS_KEY")
 AWS_STORAGE_BUCKET_NAME = os.getenv("AWS_S3_STORAGE_BUCKET_NAME")
 AWS_DEFAULT_ACL = None  # This will become default in django-storages 2.0
+LOG_BUCKET = os.getenv("LOG_BUCKET")
+AWS_LOG_KEY_BLACKLIST = ["favicon.ico"]
```
madprime: Why is this in settings.py?

open_humans/management/commands/vacuum_log_bucket.py
jhdulaney: I tried to get shlex to do [] as quotes, but it kept barfing on me.  This solution comes from here:
https://stackoverflow.com/questions/27303977/split-string-at-double-quotes-and-box-brackets

With shlex, I kept getting "ValueError: No closing quotation"

Pulling from the shlex documentation at [0]:

shlex.quotes    Characters that will be considered string quotes. The token accumulates until the same quote is encountered again (thus, different quote types protect each other as in the shell.) By default, includes ASCII single and double quotes.

Looking at the code of shlex (/usr/lib64/python3.x/shlex.py for me), it is not written to handle brackets-as-quotes.  So, in order to properly handle the datetime object, I had to find another solution.

I did consider just doing a log_entry.replace("[", '"').replace("]", '"'), but that seemed inelegant?  If you think doing that and using shlex is a better way, we can do that.


[0]  https://docs.python.org/3/library/shlex.html

open_humans/management/commands/vacuum_log_bucket.py
jhdulaney: oh, meant to uncomment that, oops.

open_humans/management/commands/vacuum_log_bucket.py
jhdulaney: That's a fair point.

open_humans/settings.py
jhdulaney: Which one?  LOG_BUCKET?  AWS_LOG_KEY_BLACKLIST?  Both?

madprime: The latter -- if it's not defined by an environment variable, and only used once in the code, I think it's clearer to define something like that it in the file that's using it (e.g. at the top)?

(I don't know if we've had a consistent approach on this, but this was my thinking when seeing it here.)

jhdulaney: Roger that, I'll move it to the module.

open_humans/management/commands/vacuum_log_bucket.py
madprime: Oops, I see, I think I'd missed that brackets were in this as well. Thanks for the clarification!

And regex's are annoying, so it's great to document that a regex came from there (rather than you creating it from scratch) -- among other things, it means one can observe if anyone else complains it's flawed, etc etc. :) Maybe you could add a comment referencing the stackoverflow?

open_humans/management/commands/vacuum_log_bucket.py
```diff
@@ -0,0 +1,112 @@
...
+
+                # Split out the key from the url
+                parsed_url = urlparse(url)
+                # parse_qs returns a dict with lists as values
+                oh_key = parse_qs(parsed_url.query).get("x-oh-key", [""])[0]
+                if oh_key == "None":
+                    oh_key = None
+                if oh_key:
+                    oh_data_file_access_logs = NewDataFileAccessLog.objects.filter(
+                        key=oh_key
```
madprime: per slack conversation, it seems like this was the reason to shift away from JSONFIELD... maybe try sticking with JSONField, I think filtering is still possible?
https://stackoverflow.com/questions/36389871/django-jsonfield-filtering

data_import/views.py
```diff
@@ -26,8 +25,11 @@ def get_and_log(self, request, key_object=None):
         """
         user = request.user if request.user.is_authenticated else None
 
-        if key_object:
-            key = {
+        access_log = NewDataFileAccessLog(
+            user=user, ip_address=get_ip(request), data_file=self.data_file
+        )
+        access_log.data_file_key = json.dumps(
```
madprime: Is this storing a string rather than a dict in the JSONField?

data_import/views.py
```diff
@@ -36,17 +38,14 @@ def get_and_log(self, request, key_object=None):
...
-            data_file=self.data_file,
-            data_file_key=json.dumps(key),
         )
-        access_log.save()
 
-        return HttpResponseRedirect(self.data_file.file_url_as_attachment)
+        aws_url = self.data_file.file_url_as_attachment
+        access_log.aws_url = aws_url
+
+        url = "{0}&x-oh-key={1}".format(aws_url, key_object.key)
```
madprime: This is minor but, if this is what's returned, should it also be what's stored in the `aws_url` as well?

data_import/views.py
jhdulaney: Oh!
Let me double check that.

PR #1015, opened 2019-03-22
Enable file download key generation globally

jhdulaney:
## Description
Previously, public data files did not have keys associated with them due to concerns over performance.  Now that the relevant API endpoints have been properly paginated, these concerns have been alleviated. This change enables file download key generations for all files, as each key points to a database entry containing much useful information.

## Related Issue
N/A

## Testing
Automated testing passes
Instrumented with Silk and found no observable difference in performance
Tested API endpoints such as the member-exchange endpoint as well as the "Your Data" section of the Activity page.

### DIFF
--- a/data_import/models.py
+++ b/data_import/models.py
@@ -115,9 +115,10 @@ class DataFile(models.Model):
     def __str__(self):
         return str("{0}:{1}:{2}").format(self.user, self.source, self.file)
 
-    @property
-    def download_url(self):
-        return full_url(reverse("data-management:datafile-download", args=(self.id,)))
+    def download_url(self, request):
+        key = self.generate_key(request)
+        url = full_url(reverse("data-management:datafile-download", args=(self.id,)))
+        return "{0}?key={1}".format(url, key)
 
     @property
     def file_url_as_attachment(self):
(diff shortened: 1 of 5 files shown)


data_import/models.py
```diff
@@ -143,13 +135,20 @@ def generate_key(self, request):
...
-                # oauth2 project auth
-                new_key.project_id = request.auth.application.id
+            if hasattr(request, "query_params"):
+                new_key.access_token = request.query_params.get("access_token", None)
+            else:
+                new_key.access_token = None
+            if hasattr(request, "auth"):
+                if hasattr(request.auth, "application"):
+                    # oauth2 project auth
+                    new_key.project_id = request.auth.application.id
```
madprime: This looks wrong, sorry I didn't catch this before.

The request.auth = an `OAuth2DataRequestProject`. This should have that project ID, which also works for the parent class `DataRequestProject`. (OnSite and OAuth2 child models share the same ID assignments, so an ID in either of those is also an ID in the parent model. Which means these are the universal way to look up any project.)

The application ID is related to the REST Framework and I don't expect it to match the IDs used by `DataRequestProject`, `OAuth2DataRequestProject` and `OnSiteDataRequestProject`. (And indeed, it doesn't.) This is going to result in mis-identifying the project that was used to download data, and we should probably simply wipe all logs before now because of the ambiguity.

madprime: (the correct thing to do is the same as onsite: `request.auth.id`)

open_humans/templatetags/utilities.py
```diff
@@ -362,3 +362,11 @@ def project_is_connected(project, user):
     Return True if the given project is connected (joined and authorized).
     """
     return project.is_joined(user)
+
+
+@register.simple_tag(takes_context=True)
```
madprime: Why did you do this instead of retaining the `property` tag on this function?

PR #1034, opened 2019-04-19
Project creation and update API

jhdulaney:
See #1035


private_sharing/api_views.py
```diff
@@ -458,3 +470,92 @@ def post(self, request):
...
+    """
+
+    authentication_classes = (CustomOAuth2Authentication,)
+    permission_classes = (HasValidProjectToken,)
+
+    def get_short_description(self, long_description):
+        """
+        Return first 139 chars of long_description plus an elipse.
+        """
+        return "{0}…".format(long_description[0:139])
```
madprime: let's only do this `if len(long_description) > 140` (otherwise just use a copy of it)

private_sharing/forms.py
```diff
@@ -127,6 +127,13 @@ class Meta:  # noqa: D101
             "deauth_webhook",
         )
 
+    def __init__(self, *args, **kwargs):
```
madprime: Is this needed? Since this is a ModelForm and this field doesn't allow blank in the database, I thought it was required already.

private_sharing/api_views.py
```diff
@@ -458,3 +470,92 @@ def post(self, request):
...
+        """
+        Take incoming json and create a project from it
+        """
+        project_creation_project = OAuth2DataRequestProject.objects.get(
+            pk=self.request.auth.pk
+        )
+
+        # If the first part of the redirect_url is provided, grab that, otherwise set
+        # to the project-creation-project's enrollment_url as a usable default
+        redirect_url_part = request.data.get("redirect-url-part", None)
```
madprime: I think it's fine to make this required.

But I don't think it's a good idea to assume it should be based on the project slug. I think that's a cool idea! But the app using this API should be able to decide whether it wants that format or not, on its end.

So the field should probably be `redirect_url`.

private_sharing/models.py
```diff
@@ -395,6 +395,7 @@ class Meta:  # noqa: D101
             "/direct-sharing/oauth2-setup/#setup-oauth2-authorization"
         ),
         verbose_name="Redirect URL",
+        blank=True,
```
madprime: Is this needed? Or maybe it's leftover from realizing we needed to specify the redirect_url...

private_sharing/models.py
```diff
@@ -408,6 +409,8 @@ class Meta:  # noqa: D101
         verbose_name="Deauthorization Webhook URL",
     )
 
+    diy_project = models.BooleanField(default=False)
```
madprime: Minor, but I'd prefer diyexperiment to be consistent with the app we've started making. :) (OK if we change the name later, but we should be consistent about using the current working title.)

private_sharing/api_views.py
```diff
@@ -458,3 +470,92 @@ def post(self, request):
...
+
+            # Generate redirect URL and save to project
+            project.redirect_url = "{0}/{1}/complete/".format(
+                redirect_url_part, project.slug
+            )
+            project.save()
+
+            # Serialize project data for response
+            # Copy data dict so that we can easily append fields
+            serialized_project = ProjectDataSerializer(project).data
```
madprime: This is redundant with the project data endpoint, which is OK -- but it's also insufficient.

That serializer does not include client_id or client_secret -- and it can't include the latter, for security reasons. The API needs to also return both of those things -- otherwise the other app can't manage the project on the coordinator's behalf, going forward.

private_sharing/api_authentication.py
```diff
@@ -13,6 +18,37 @@
...
+def make_oauth2_tokens(project, user):
+    """
+    Returns a tuple, an AccessToken object and a RefreshToken object given a project and a user.
+    :param project:  An oath2 project
+    :param user:  The user for the access token and refresh token
+    If project is not a valid oauth2datarequestproject, returns None
+    """
+    if not project.__class__ == OAuth2DataRequestProject:
+        return None
+    expires = (
```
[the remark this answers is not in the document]
jhdulaney: Probably not?  I'll give that a try

private_sharing/models.py
jhdulaney: It's a leftover from having to specify a redirect_url.

private_sharing/api_views.py
jhdulaney: Oh, I'd had those in my mockup, but somehow missed including them here.  Oops!

private_sharing/forms.py
jhdulaney: Artifact of the redirect_url change, reverting.

private_sharing/api_authentication.py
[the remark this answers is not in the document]
jhdulaney: So, two ways to do the same thing, but, for consistency, I switched to the way oauth-toolkit does the thing.

private_sharing/api_views.py
```diff
@@ -458,3 +470,102 @@ def post(self, request):
...
+                access_token, refresh_token = make_oauth2_tokens(project, member.user)
+                serialized_project["coordinator_access_token"] = access_token.token
+                serialized_project["coordinator_refresh_token"] = refresh_token.token
+
+            return Response(serialized_project, status=status.HTTP_201_CREATED)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+class ProjectUpdateAPIView(APIView):
```
madprime: OAuth2ProjectUpdateAPIView

private_sharing/api_views.py
```diff
@@ -458,3 +470,102 @@ def post(self, request):
...
+    """
+    API Endpoint to update a project.
```
madprime: ... to update an OAuth2DataRequestProject

private_sharing/api_views.py
```diff
@@ -458,3 +470,102 @@ def post(self, request):
...
+                add_data=False,
+                explore_share=False,
+                active=True,
+                short_description=self.get_short_description(
+                    serializer.validated_data["long_description"]
+                ),
+                coordinator=member,
+                leader=member.name,
+                request_username_access=False,
+                diyexperiment=True,
```
madprime: this should default to False and be True if the parameter was provided and is true :)

private_sharing/api_views.py
```diff
@@ -458,3 +470,102 @@ def post(self, request):
...
+            return "{0}…".format(long_description[0:139])
+        return long_description
+
+    def post(self, request):
+        """
+        Take incoming json and create a project from it
+        """
+        member = get_oauth2_member(request).member
+        serializer = ProjectAPISerializer(data=request.data)
+        if serializer.is_valid():
```
madprime: Our other APIs use re-use the on site forms for validation: https://github.com/OpenHumans/open-humans/blob/master/open_humans/api_views.py

I think we should be doing the same here, to have consistency in validation (i.e. re-using `OAuth2DataRequestProjectForm`)

PR #1041, opened 2019-04-29
Admin page speedups

jhdulaney:
## Description
Some of the admin pages are slow because they load the full member (or user) list.  This corrects that slowness.

## Related Issue
N/A

## Testing
Verified that DataRequestProject admin pages load much faster now

### DIFF
--- a/open_humans/admin.py
+++ b/open_humans/admin.py
@@ -10,6 +10,7 @@ class MemberAdmin(admin.ModelAdmin):
     """
 
     raw_id_fields = ("user",)
+    search_fields = ("user__username", "user__email")
 
     def get_queryset(self, request):
         """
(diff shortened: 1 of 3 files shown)


jhdulaney:
> I've suggested an addition via a PR to your branch that makes it a _lot_ easier to find Members (it's pretty impossible to find and change the "coordinator" in the current admin).
> 
> It seems like the important fix here was `raw_id_fields`? Is `get_queryset` necessary? (It doesn't seem important when I test this locally?)

Profiling with silk seems to indicate an 80 ms difference.  I guess since it's the admin site, that's not enough difference to warrant it?

madprime:
@jhdulaney yeah, I think I'd prefer to keep it simpler, that's a pretty minor gain for an admin page :)

jhdulaney:
Pushed an update

PR #1057, opened 2019-05-28
rm legacy approach in public sharing model

madprime:
## Description
This replaces a legacy approach to managing public data sharing, which has been using a "source label" approach. The model now refers directly to the project membership as a `OneToOneField`.

This is likely to change again if/when we work out how to use the new `DataTypes` in managing public data sharing.

This also makes some updates to `DataFile` and `ProjectDataFile`. (As part of this work I cleaned up some weirdnesses here as well. The presence of two models is a legacy behavior, as now every `DataFile` object should also have a `ProjectDataFile`.)

## Related Issue
This legacy approach in public data sharing status was encountered as I created a rewrite of the activity pages.

## Testing
  * passed automated testing locally (and updated existing testing to fix places it broke)
    * load activity pages
    * turn public data sharing off and on
    * ... and saw files disappear/appear in the API and profile page
    * quit/remove data from a project
    * list projects on "My data" page & am able to toggle there as well


data_import/models.py
```diff
@@ -159,7 +129,7 @@ def generate_key(self, request):
 
     @property
     def is_public(self):
-        return is_public(self.user.member, self.source)
```
madprime: I updated DataFile & ProjectDataFile because of this. The updated models now use a database query rather than the circuitous "look up from a string" approach. That approach is still in the code because a couple other places still use it, but hopefully on the way to full removal. ;)

data_import/models.py
```diff
@@ -81,29 +69,11 @@ class DataFileManager(models.Manager):
...
-
     def contribute_to_class(self, model, name):
         super(DataFileManager, self).contribute_to_class(model, name)
 
         models.signals.pre_delete.connect(delete_file, model)
 
-    def public(self):
-        prefix = "user__member__public_data_participant__publicdataaccess"
-
-        filters = {prefix + "__is_public": True, prefix + "__data_source": F("source")}
```
madprime: This was moved to ProjectDataFile and switched away from the "data source" approach (as this was removed from PublicDataAccess).

data_import/models.py
```diff
@@ -81,29 +69,11 @@ class DataFileManager(models.Manager):
     pre_delete signal connected correctly.
     """
 
-    def for_user(self, user):
```
madprime: Moved to ProjectDataFile as a more natural place for it, since this manager was filtering on a field of that model.

PR #1063, opened 2019-07-16
Public APIs Overhaul

### LINKED ISSUE #1033: fix users-by-source API endpoint
Users by source public data API endpoint currently lists projects that aren't data sources, and fails to list projects that are no longer active.

madprime:
## Description
This creates a new set of publicly available APIs that present public information/data about the following:
  - DataFiles
  - DataTypes
  - Members
  - Projects
    
The naming schemes and formats are generally re-used, and nesting has been avoided -- for example, project serialization doesn't list public files for each project, that's available at another endpoint. (Nesting has caused previous endpoints to crash.)

Documentation is updated as well.
    
Some previous endpoints are retained, as they are known to be in use, or might be. These are deprecated.

Django version is also bumped, and a model was updated to add a related_name for reverse lookups.

## Related Issue
This closes #1033 by rendering that endpoint obsolete.

## Testing
  * added automated testing for the top level "list" endpoints
  * ran locally to test all endpoints manually
  * loaded code into staging for further testing and viewing


open_humans/templates/pages/public-data-api.html
```diff
@@ -15,162 +15,443 @@
...
+              href="/api/public-data/?username=madprime&limit=5">
+                /api/public-data/?username=madprime&limit=5</a>
+            </code>
+          </p>
 
-    <p>
-      <strong>URL:</strong> <span id="source-url"></span>
-    </p>
+          <p>
+            This query shows all Harvard Personal Genomes Project data updated on
```
gedankenstuecke: Should we maybe use another example? The PGP isn’t listed publicly right now as it’s inactive, isn’t it? Makes it harder for people to understand what’s going on?

PR #1067, opened 2019-09-04
Expand private project datafile serialization

madprime:
This adds "datatypes" and "source_project" fields to the datafile serialization in private project APIs, matching the equivalent fields in the [public API datafile serialization](https://www.openhumans.org/api/public/datafiles/).

## Description

As part of making DataTypes useful in the site, I'd like to make them usable by our Jupyter Notebooks servers – but I noticed that the `datatypes` field is _not_ available in private project API datafile serialization. It _is_ available in the public datafile API datafile serialization.

This update adds two new fields to the private project API serialization to match the public API fields: `datatypes` and `source_project`. The second is arguably redundant with `source`, but possibly more useful than the old-style "direct-sharing-" label.

Both fields are providing API paths as identifiers rather than bare IDs. See examples below.

## Examples

Current API behavior returns:
```
{
  'id': 50988,
  'basename': '23andMe-genotyping.txt',
  'created': '2018-05-11T02:46:25.300588Z',
  'download_url': 'https://www.openhumans.org/data-management/datafile-download/50988/?key=32372de0-e355-4c01-8be0-ca117aa51b91',
  'metadata': {
    'tags': ['23andMe', 'genotyping'],
    'description': '23andMe full genotyping data, original format',
    'creation_date': '2018-05-11 02:46:25+00:00'},
  'source': 'direct-sharing-128'
}
```

Updated behavior (new lines bold):
```
{
  'id': 50988,
  'basename': '23andMe-genotyping.txt',
  'created': '2018-05-11T02:46:25.300588Z',
  'datatypes': ['/api/public/datatype/13/'],
  'download_url': 'https://www.openhumans.org/data-management/datafile-download/50988/?key=32372de0-e355-4c01-8be0-ca117aa51b91',
  'metadata': {
    'tags': ['23andMe', 'genotyping'],
    'description': '23andMe full genotyping data, original format',
    'creation_date': '2018-05-11 02:46:25+00:00'},
  'source': 'direct-sharing-128',
  'source_project': '/api/public/project/128/'
}
```

## Testing

Tried this out with a local version of the site. Passed automated tests.

### DIFF
--- a/data_import/serializers.py
+++ b/data_import/serializers.py
@@ -1,10 +1,9 @@
 from collections import OrderedDict
-from urllib.parse import urlparse, parse_qs
 
 from django.urls import reverse
 from rest_framework import serializers
 
-from private_sharing.models import DataRequestProject
+from common.utils import full_url
 
 from .models import AWSDataFileAccessLog, DataFile, DataType, NewDataFileAccessLog
 
@@ -55,14 +54,16 @@ class DataFileSerializer(serializers.Serializer):
         Get links to DataType API endpoints for file DataTypes
         """
         return [
-            reverse("api:datatype", kwargs={"pk": dt.id})
+            full_url(reverse("api:datatype", kwargs={"pk": dt.id}))
             for dt in obj.parent_project_data_file.datatypes.all()
         ]
 
     def get_source_project(self, obj):
-        return reverse(
-            "api:project",
-            kwargs={"pk": obj.parent_project_data_file.direct_sharing_project.id},
+        return full_url(
+            reverse(
+                "api:project",
+                kwargs={"pk": obj.parent_project_data_file.direct_sharing_project.id},
+            )
         )
 
 
--- a/data_import/serializers.py
+++ b/data_import/serializers.py
@@ -42,11 +42,28 @@ class DataFileSerializer(serializers.Serializer):
         ret["id"] = instance.id
         ret["basename"] = instance.basename
         ret["created"] = instance.created
+        ret["datatypes"] = self.get_file_datatypes(instance)
         ret["download_url"] = instance.download_url(request)
         ret["metadata"] = instance.metadata
         ret["source"] = instance.source
+        ret["source_project"] = self.get_source_project(instance)
 
         return ret
+
+    def get_file_datatypes(self, obj):
+        """
+        Get links to DataType API endpoints for file DataTypes
+        """
+        return [
+            reverse("api:datatype", kwargs={"pk": dt.id})
+            for dt in obj.parent_project_data_file.datatypes.all()
+        ]
+
+    def get_source_project(self, obj):
+        return reverse(
+            "api:project",
+            kwargs={"pk": obj.parent_project_data_file.direct_sharing_project.id},
+        )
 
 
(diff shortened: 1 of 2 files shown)


madprime:
@gedankenstuecke the main question I have – and would like to hear what you think – is whether you think having the paths is the correct choice, rather than bare IDs (or some other option)? That is, `/api/public/datatype/13` rather than `13`.

beaugunderson:
there's precedence for using URIs as identifiers, for example in XML DTDs (though I'd probably include the base URL too)

data_import/serializers.py
```diff
@@ -42,12 +42,29 @@ def to_representation(self, instance):
...
         ret["source"] = instance.source
+        ret["source_project"] = self.get_source_project(instance)
 
         return ret
 
+    def get_file_datatypes(self, obj):
+        """
+        Get links to DataType API endpoints for file DataTypes
+        """
+        return [
```
gedankenstuecke: As @beaugunderson mentioned, it would be good to give the full URL including the `BASE_URL` here to make using those endpoints easier. Also helps making it more consistent as the `download_url` return is already giving the full URL as well.

data_import/serializers.py
```diff
@@ -42,12 +42,29 @@ def to_representation(self, instance):
...
+            reverse("api:datatype", kwargs={"pk": dt.id})
+            for dt in obj.parent_project_data_file.datatypes.all()
+        ]
+
+    def get_source_project(self, obj):
+        return reverse(
+            "api:project",
+            kwargs={"pk": obj.parent_project_data_file.direct_sharing_project.id},
```
gedankenstuecke: see comment above.

madprime:
Thanks @beaugunderson – yeah, in particular the URIs as identifiers was something I'd seen with JSON LD. (I didn't implement JSON LD, I think that's still possible someday. IIRC the use of `@` leading fields might've made it weird/tricky for Django REST Framework to support.)

Whether to do relative or absolute wasn't clear to me – the internet doesn't seem to have strong opinions on this point that I found. https://webmasters.stackexchange.com/questions/99537/absolute-or-relative-local-url-in-schema-org-json-ld

@gedankenstuecke you _did_ have a chance to say something on the public data API & didn't say anything about full vs relative URLs... :stuck_out_tongue:  https://github.com/OpenHumans/open-humans/pull/1063

**Can you weigh in on whether you'd like to see the public data API also switch to absolute URLs in all these cases?**

This probably won't break anything, as it was only recently updated. If you think we should do it this way, I think it should be done there as well; I'd add that to this PR.

gedankenstuecke:
My bad for not chiming in at the public data API, I guess @beaugunderson just gave me the idea now and as someone coding against the APIs I would find the full URLs more comfortable.  

So yes, I think it would be good to update the public data API as well to make it easier for folks!

madprime:
@gedankenstuecke ok, sounds good – I'll update that too. I didn't have a strong feeling either way, probably the relative URL was just easier – you made a good point re the `download_url` and more consistency with that. :)

madprime:
Going ahead with merge & deploy for this. Thanks!

PR #1069, opened 2019-09-09
Enable easier management of allowed oauth2 redirects

madprime:


madprime:
fyi for @gedankenstuecke – I'm moving ahead with merging this so I can respond to the support request ASAP for the project user that contacted us. :)

This should make it easier to add more allowed URLs via an environment variable but it's still pretty sucky.

madprime:
(this also includes a super minor change to the bulk email management command)

PR #1078, opened 2019-11-05
DataType features

madprime:
## Description
Adds new DataType fields and features to better enable user-defined uploads.

This adds the following fields:
* `uploadable` – boolean, True if the DataType is considered uploadable by users
* `details` – textfield, expected to be Markdown format

This also adds a new Bootstrap 4 base and converts all the DataType pages to use this base. The goal is a gradual transition to BS4 in the site to take advantage of its features; if there are bugs/issues with it, they should only occur in these pages (and not the full site).

This adds collapse to "history" elements when viewing information for a specific DataType (as they're much larger with the added "details" field).

This also adds these fields to the public API, including a boolean filter on the new "uploadable" field.

This removes listing of unapproved project sources on a DataType, see #1068 

Open questions regarding DataType behavior:
1. Should the `description` field be shorter? Currently our limit is 512 characters but the longest in use is 74 characters. The `details` field is Markdown formatted and more appropriate for extended descriptions.
2. When should a DataType be editable? Currently these are only editable as long as there are no approved projects that have registered them as a data source. (As soon as an approved project has registered as a source, they become frozen/uneditable.) In particular, this is an anticipated issue for "uploadable" DataTypes which may need ongoing updates to instructions for collecting and retrieving data.

## Testing
    * load all updated DataType pages
    * use the create form
    * use the update form
    * try logging in via username/password & logging out


gedankenstuecke:
> Should the description field be shorter? Currently our limit is 512 characters but the longest in use is 74 characters. The details field is Markdown formatted and more appropriate for extended descriptions.

Good question. Let's go for the traditional 140 characters on this? Seems like a number we've all been trained to use for a long while that's a good compromise of length and conciseness? 

> When should a DataType be editable? Currently these are only editable as long as there are no approved projects that have registered them as a data source. (As soon as an approved project has registered as a source, they become frozen/uneditable.) In particular, this is an anticipated issue for "uploadable" DataTypes which may need ongoing updates to instructions for collecting and retrieving data.

How do we feel about the markdown `description` field being editable even after approval? Especially how to get the data can change at any point (e.g. I had to update the Twitter Archive Analyzer recently to adapt to the new layout of Twitter). Alternatively, maybe we need a `How to get this data` field that remains editable?

madprime:
Updated to add the character limits and new behavior for locking editing. Also added enforcement of "terminal" (no children) status for any uploadable datatypes.

PR #1092, opened 2020-01-29
Update about page

madprime:


madprime:
Going ahead with merge so this can be reviewing in staging.

PR #1094, opened 2020-02-01
Update welcome email

madprime:
…among other things it corrects the "mission" to match the mission statement of the organization. Further improvements are probably possible, but I think the current version is unacceptably dated and fails to reflect the focus we've moved to in the last couple years.

### DIFF
--- a/open_humans/templates/email/welcome.html
+++ b/open_humans/templates/email/welcome.html
@@ -1,10 +1,8 @@
 {% load utilities %}<p><strong>Thank you for joining the Open Humans community!</strong></p>
 
 <p>
-  It's our mission to help connect you and your data with researchers.
-  In Open Humans, you'll be able to build a profile with data you receive
-  from studies and commercial sources. We'll invite you to share that data
-  in various ways, publicly or directly with projects.
+  It's our mission to empower you with your personal data, to explore and
+  share for the purposes of education, health, and research.
 </p>
 
 <p>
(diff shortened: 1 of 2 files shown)


open_humans/templates/email/welcome.html
```diff
@@ -15,13 +13,13 @@
...
 
 <p>
-  We have many data sources you can add! Each source you add is a generous
-  contribution to science and gets you a badge on your public profile. It also
-  helps enable more studies because researchers can design studies more easily
-  when they know members have data ready to share.
+  Your account can be used to collect and manage any personal data you
+  choose to connect to Open Humans. We have activities created by us, as
+  well as other members, that enable adding data from a variety of sources.
+  enable adding data from a variety of sources.
```
beaugunderson: @madprime copy paste issue? :)

open_humans/templates/email/welcome.html
```diff
@@ -1,10 +1,8 @@
 {% load utilities %}<p><strong>Thank you for joining the Open Humans community!</strong></p>
 
 <p>
-  It's our mission to help connect you and your data with researchers.
-  In Open Humans, you'll be able to build a profile with data you receive
-  from studies and commercial sources. We'll invite you to share that data
-  in various ways, publicly or directly with projects.
+  It's our mission to empower you with your personal data, to explore and
```
beaugunderson: i think this could be a little clearer; the comma after personal data trips me up (what am i exploring and sharing?)

open_humans/templates/email/welcome.html
madprime: Whoops! Not sure how it happened, I guess I was tired – good catch, thank you!

open_humans/templates/email/welcome.html
madprime: Fair point. But... I think I'd like to leave it vague for now. Hoping the board will be able to discuss more at our next meeting, we may be moving in directions that aren't requiring sharing data itself (i.e. sharing in other ways: tools, notebooks, other collaborating).

madprime:
Leaving here for the record – @gedankenstuecke said this seemed good & I'd like to merge so it's updated sooner rather than later (further improvements always possible). Thank you to @beaugunderson for the review, it really helps to have other eyes on it!

PR #1096, opened 2020-02-10
Add change username form

madprime:
## Description
Give people a form to change their username themselves. Asking us to change it was a legacy issue.

## Testing
    * confirmed username change works
    * confirmed error on existing username
    * confirmed custom validation (error on disallowed characters)

### DIFF
--- a/open_humans/forms.py
+++ b/open_humans/forms.py
@@ -2,10 +2,12 @@ from captcha.fields import ReCaptchaField
 
 from django import forms
 from django.conf import settings
+from django.contrib.auth import get_user_model
 from django.core.mail import send_mail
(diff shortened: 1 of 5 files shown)


gedankenstuecke:
Do you have an idea how this will interact with Discourse, where the OH username is used? I noticed that users can't change their discourse username manually, so will it propagate from OH in this case?

madprime:
@gedankenstuecke I think it should "just work". Actually, I've turned on some settings in the forums; it used to override username & avatar, but not name & email (I think it should do both – and I wonder if that would've fixed the new-email-new-account issue we had earlier...)

image

gedankenstuecke:
Ok, cool!

PR #1107, opened 2020-04-16
[MRG] add apple oauth

madprime:
Work in progress, not ready to merge. Current status:

* button created for login
* trying an open PR adding this feature in django-allauth
* managing button display with a django-waffle feature flag


madprime:
Tested and was able to login! :tada: I'll let @gedankenstuecke try it out too (see Slack DM for instructions on how to test).

I think this can be merged to production, as the button is controlled by a django-waffle flag. We can test it there, get it working, then switch it on for everyone (and later remove waffle-specific code).

gedankenstuecke:
Ok, I gave it a try and ran into an error during the signup. I used my Apple ID and went with sharing my email address. It took me to the account creation page (username + display name). When trying to submit this form I got `ValidationError: ["'true' value must be either True or False."]`

The full sentry log for this is at https://sentry.io/organizations/openhumansorg/issues/1616328235/

Now I also end up with my apple-ID being in a weird state, as retrying leads to 

```
DoesNotExist at /account/social/signup/
EmailAddress matching query does not exist.
```

So some remnants of my first try have been saved. I tried fixing it for testing by deleting the social account, but even that lead to the same `DoesNotExist` error as above.

gedankenstuecke:
Okay, figured out that I needed to also delete my existing `User` object that was created in the failed attempt to give it another try. But: Even after deleting all information and creating a new account the error remains the same. I assume this is a consequence of my apple-ID email not having an account yet.

madprime:
This has been updated to use the [recent PR fixing issues](https://github.com/0leg5ergeev/django-allauth/pull/11) added by @mhsparks to the ongoing [PR to add Apple OAuth to django-allauth](https://github.com/pennersr/django-allauth/pull/2424).

This fixes various issues we encountered when testing. As far as we can tell, testing in staging, this seems to work. I switched to using a copy of this version of `django-allauth` that's in my own repo, to be sure it doesn't disappear.

The new feature remains "hidden" via a django-waffle feature flag, and we're likely to keep it that way until the Apple login feature here is merged to `master` in `django-allauth` (:crossed_fingers:!) Hopefully that does happen, and in nearly the same state – I hope I'm not faced with cleaning up the database, but I think we can handle that if needed.

Right now, there's urgency in adding this to enable a mobile app supporting Apple Watch & Apple Health data import, for Quantified Flu (which relates to the coronavirus crisis). After discussion with @gedankenstuecke, we felt it was safe to merge to production to unblock that effort, especially if the feature remains "hidden".

madprime:
Update to add Apple OAuth login

- uses an in-progress PR branch for django-allauth, hoping that this is what is eventually merged into master
- the feature currently remains "hidden" via an django-waffle feature flag

PR #1109, opened 2020-05-08
Custom django-waffle middleware to store testing flags

madprime:
## Description
This makes django-waffle testing flags "stick" when not used by an initial view or template.

django-waffle has an open issue regarding testing flags not being stored in cookie unless there's some attempt to use them in a view or template: https://github.com/django-waffle/django-waffle/issues/80

While the issue seems to think this isn't solvable without breaking backwards compatibility, it seems like my modification to the middleware should generally work? (It uses the flag customization that was described as an issue.)

So, this seems to work for me, but maybe I'm missing something. (And I might submit this solution as a PR to django-waffle.)

## Testing

* Manually tested that a waffle testing flag is now set in a request/response that previously lost the flag (our login redirect)
* Passed automated testing

### DIFF
--- a/open_humans/middleware.py
+++ b/open_humans/middleware.py
@@ -5,9 +5,15 @@ from urllib.parse import urljoin
 from django.conf import settings
 from django.http import HttpResponseRedirect
 
+from waffle import get_waffle_flag_model
+from waffle.middleware import WaffleMiddleware
+from waffle.utils import get_setting as get_waffle_setting
+
 from .models import Member
 
 logger = logging.getLogger(__name__)
+
+WaffleFlag = get_waffle_flag_model()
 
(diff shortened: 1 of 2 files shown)


dragon-dxw:
@madprime Could you chime in on https://github.com/jazzband/django-waffle/pull/505 and confirm that you're happy for your code to be used in the project in this way? And possibly any problems that implementing these changes might cause?

gedankenstuecke:
hey @dragon-dxw, at least we didn't run into any issues with it, but also have to say we didn't end up using it a whole lot on our end! 

Regarding using the code: I can't speak for @madprime, but would assume that yes, as our code base is MIT licensed 🙂

PR #1111, opened 2020-05-13
Change password form fixes and clean up

### LINKED ISSUE #1110: Set password form is missing
Reported by @mavaddat (thank you!): the "set password" form is missing/broken.

This form is used when a password doesn't already exist (otherwise it's the "change password" form) – which only happens when someone creates an account via social login, _then_ adds a password.

Because we have custom overrides for the django-allauth templates, it was trying to load a default template (which was broken). (And because we added social login after username/password, we had missed this form was now needed.)

madprime:
Fix missing "set password" form and clean up code.

## Description

Code clean-up is to reduce customization/override on django-allauth. In particular, I wanted to remove custom password validation, since this is now built in to Django since v2.0 (i.e. "minimum length" is now set in `settings.py`).

The missing "set password" form is due to how social login ends up needing this, for new accounts created via social login. When we had just username/password, this form would never be triggered, only the "change password" form (because a user would always already have a password).

## Related Issue
Fixes #1110 

## Testing

  * updated automated testing to reflect the behavior of django-allauth on password resets
    * reproduced a "no password" status and tested the "set password" form
    * tested "set password", "change password", and "reset password" forms all worked after updates


madprime:
Going ahead with merge, we can double-check everything in staging.

PR #1115, opened 2020-07-01
Add optional signature for webhooks

madprime:
## Description

This update:

* adds an optional webhook_secret field an OAuth2 project can provide
* if provided, this is used to create an hmac-sha1 hexdigest to verify
a request made to a webhook
* this is included as a custom header "X-OpenHumans-Webhooks-Signature"
when calling a webhook
* documentation is added for the deauthorization webhook and webhook secret
* the JSON payload for the deauthorization webhook has been corrected
(it was accidentally doing a second JSON encoding)

The last item will break anything that was using this webhook. However,
the webhook wasn't actually documented before now. Support for it was added
to the django-open-humans library, and so a concurrent update to that
package is staged.

## Testing

  * manual testing via setting up a local OAuth2 project instance to verify:
    * registering the webhook secret
    * reproducing and verifying the signature for a given payload
    * absence of header when webhook secret is not provided

An update to the django-open-humans package was concurrently developed as part of this testing: https://github.com/OpenHumans/django-open-humans/pull/21


madprime:
Because removing the accidental double-encoding of JSON is likely to break anything that was using this webhook, I've updated the PR to add a fallback behavior: if the POST receives a non-200 response, the server retries with the old double-encoded JSON.

I tested with django-open-humans, which returns a server error (500) and then a 200 after receiving the fallback (double-encoded) version.

@gedankenstuecke if this all looks good to you, I'd like to merge and deploy :)

PR #1125, opened 2020-12-22
Attempt to fix API error

madprime:
Based on server errors, this part of the code seems to be the issue.

This update attempts to return the code to a prior state:
https://github.com/OpenHumans/open-humans/blob/a3f31da0c98c4b27bb1522c1e38b1ceb3a3e83e3/data_import/models.py#L119


gedankenstuecke (approved):
Okay, that makes sense! Let's give it a try!

PR #1128, opened 2020-12-22
Set primary key to UUID to fix API bug

madprime:
While objects were routinely cleaned, the auto-incrementing ID field
went past the limit of IntegerField. The UUID field is unique and can
be used instead.

## Testing

After fumbling I finally reproduced the issues locally. The "id" field was at fault.

In local testing, I didn't see any issues with this migration, which removes that field and
uses the UUID as primary key. That said, I only manually tested the exchange member endpoint.


madprime:
Going ahead with a merge and deploy as this is an urgent issue.

abitrolly:
Using UUIDs as a primaty key may sound as bad idea for all performance reasons, but for this specific use case it seems okay.

PR #1147, opened 2021-07-22
Update requirements files with newer pip-tools

madprime:
Because of how pip-tools handles formatting, this looks like a
major change but is actually just "moving comments around".


madprime:
@gedankenstuecke I scanned pretty thoroughly... as far as I can tell, there's no "actual" changes. Just a change in formatting. (And there's no option to use "old style", but of course there is an issue requesting it... 😂 )

abitrolly:
@madprime link the issue?

gedankenstuecke (approved):

madprime:
@abitrolly If you're wondering what the pip-tools issue I referred to was, I think it was this one: https://github.com/jazzband/pip-tools/issues/1306

I'm happy to switch over to the new formatting (I'm not sure I'd use this option if it were available), but I can see from the discussion there that there are various reasons folks would prefer to have the old style option still available. :)

PR #1150, opened 2021-08-17
Switching  to avoid leading zeros in future memberID creations

danamlewis:
## Description
Previously memberIDs could be created with 0 in the front. Various programs for data processing drop the 0, and then re-using the project memberIDs back in the command line would cause various errors in pulling or processing the data. Immense headaches have ensued for project administrators and data users. This attempts to remove the leading zeros (although it doesn't solve existing memberIDs with leading zeros) from future projectmemberID creation.

## Testing
Have not tested but used OpenAI Codex to validate the sanity of this potential solution.

### DIFF
--- a/open_humans/models.py
+++ b/open_humans/models.py
@@ -42,12 +42,14 @@ def get_grant_project_image_upload_path(instance, filename):
 
 def random_member_id():
     """
-    Return a zero-padded string from 00000000 to 99999999 that's not in use by
-    any Member.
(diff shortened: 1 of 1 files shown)


abitrolly:
Why not switch to strings then?

beaugunderson:
The code looks good; I'm curious if this implies a secondary step to re-assign user IDs that currently have leading zeroes? If so, this code helps limit the number of reassignments that will happen if we do take that step in the future so it's worth merging now reagrdless.

